PR #8, opened 2019-10-18
BCDA-2005 Maintenance: DB Schema Migration Strategy

dhgreene:
### Fixes [BCDA-2005](https://jira.cms.gov/browse/BCDA-2005)
Up to this point, SSAS has been using GORM automigration to maintain the database schema.  This is a somewhat opaque, one-way process that works well in development environments, but is less desirable in production.

This PR introduces the database schema migration tool `golang/migrate`, which uses SQL scripts to make changes and to roll them back.

### Proposed Changes
- Call `golang/migrate` instead of GORM automigrate for all database schema changes
- Begin writing SQL scripts to make changes and roll them back
- Begin testing SQL migrations as part of `make test`

### Change Details
- A database dump of the current SSAS tables/relations/indices constitutes the first migration
- `make migrations-test` runs migration tests, and is run at the end of `make test`

### Security Implications
A new executable (`migrate`) is added to local containers that need to make database schema changes.  How to use this tool in AWS environments will be covered by a separate PR.

- [x] security checklist is completed for this change
[Security Checklist - SSAS DB Schema Change Strategy](https://confluence.cms.gov/display/BCDA/Security+Checklist+-+SSAS+DB+Schema+Change+Strategy)

### Acceptance Validation
Screen Shot 2019-10-15 at 10 53 21 AM

### Feedback Requested
- Actively soliciting optimizations to the Docker images and invocations
- Is the testing strategy sound?
- Does the README.md need more content?  What kind?
- All other comments welcome


ops/migrations_test.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# This script is intended to be run from within the Docker "unit_test" container
+# The docker-compose file brings forward the env vars: DB
+#
+set -e
+set -o pipefail
+
+timestamp=`date +%Y-%m-%d_%H-%M-%S`
+
+echo "Setting up test DB (bcda_test)..."
```
msnook: Should the test db be named `ssas_test` in this env?

unit_test.sh
```diff
@@ -18,9 +18,9 @@ echo "TEST_DB_URL is $TEST_DB_URL"
 usql $DB_HOST_URL -c 'drop database if exists bcda_test;'
 usql $DB_HOST_URL -c 'create database bcda_test;'
```
msnook: `ssas_test` instead of `bcda_test`?

Dockerfiles/Dockerfile.tests
```diff
@@ -13,6 +13,12 @@ COPY . .
 RUN dep ensure
 
 RUN apt-get update
+RUN apt-get install apt-transport-https ca-certificates libunwind8 icu-devtools python-pip -y
```
msnook: Does `golang-migrate` need to exist in the test image?  Could the `migrations_test.sh` simply be triggered via CLI from the ssas container?

Makefile
```diff
@@ -29,6 +29,9 @@ postman:
 	# For example: make postman env=test token=<MY_TOKEN>
 	docker-compose -f docker-compose.test.yml run --rm postman_test test/postman_test/SSAS.postman_collection.json -e test/postman_test/ssas-local.postman_environment.json --global-var adminClientId=$(SSAS_ADMIN_CLIENT_ID) --global-var adminClientSecret=$(SSAS_ADMIN_CLIENT_SECRET)
 
+migrations-test:
+	docker-compose -f docker-compose.test.yml run --rm tests bash ops/migrations_test.sh
```
msnook: Could this just be run via CLI from the SSAS container?  Is it pointed at the same DB in both cases?  If so, I would argue that `golang-migrate` can be removed from the tests image.

db/migrations/README.md
```diff
@@ -0,0 +1,22 @@
+# Database Schema Changes
+
+## Philosophy
+* All database changes will be modeled in discrete schema migrations
+    * Code to roll back a migration will be included unless highly impracticable
+* Migrations will receive unique sequential integer identifiers.
+    * Engineers should keep aware of identifier collisions (and in some instances required migration order), and adjust their migration ID’s before merging with master, if necessary.
```
msnook: If a migration is run on `0001` and a few days later `0002` is written, what happens if I run `up`?  Will both `0001` and `0002` be executed?  Or is it smart enough to know that only `0002` needs to be run?

msnook:
Good stuff.  I'm curious how we will transition to using this in AWS.  We cannot start at `0001` in AWS, since we already have our DB in place.

ops/migrations_test.sh
dhgreene: Probably.  But the main database is still `bcda`, so I'm not sure whether `bcda_test` is more congruent.

Dockerfiles/Dockerfile.tests
dhgreene: It needs to be there for the migrations test, which does not run in the ssas container.

Makefile
dhgreene: No, the SSAS container points at the non-test DB.

db/migrations/README.md
dhgreene: It's smart enough to know, by keeping track in a table named `database_migrations`.

ops/migrations_test.sh
msnook: Yeah, makes sense to leave it as is.

dhgreene:
@msnook I've made changes based on our conversation on Friday, and this is ready for re-review.

PR #9, opened 2019-10-25
BCDA-1984 Refactor: Allow groups to share group_id with deleted groups

dhgreene:
### Fixes [BCDA-1983](https://jira.cms.gov/browse/BCDA-1983) and [BCDA-1984](https://jira.cms.gov/browse/BCDA-1984)

The `groups` table requires each group to have a unique group_id (a string identifier specified by the API user when creating a group).  Because we are soft-deleting groups (setting `deleted_at` instead of removing the group), this means that a group_id cannot be reused if a "deleted" group is using it.  😱 

Fixing this problem takes two changes (thus two tickets):
- Change the uniqueness requirement on `groups.group_id` to an index that can ignore soft-deleted groups
- Key the `systems` table off of `groups.id` rather than `groups.group_id`

### Proposed Changes
- Add a column `systems.g_id` to reference `groups.id`.  
- Leave the existing `systems.group_id` to reference `groups.group_id`.

### Migration Strategy
1. Run the first .sql script (schema migration) on the database before deploying the code.  This sets up the database to support either the existing code, or the new changes by:
   - Adding the new column `systems.g_id`
   - Running a script to populate existing systems' `g_id` based on their `group_id`
   - Adding a trigger to automatically populate the `g_id` for new/updated systems
1. If all is well, deploy the code.  Otherwise, run the rollback script.
1. Assuming we've gotten this far, run the second .sql script to clean up the trigger.

### Security Implications
This is a reorganization of existing information and functionality, and should have no effect on security.

### Acceptance Validation
The unit tests for the migration 🙌 test making the database schema changes and reverting them:
Screen Shot 2019-10-25 at 12 35 21 AM

### Feedback Requested
- Is having two references from `systems` to `groups` unwise, or is it an acceptable compromise between normalization and reducing lookup queries?
- Is there a better name than `systems.g_id`?
- What else can be improved?


msnook:
Can you merge `master` into this branch to simplify the review process?

PR #11, opened 2019-11-04
BCDA-2225 Docker image streamlining for DB migration

dhgreene:
### Fixes [BCDA-2225](https://jira.cms.gov/browse/BCDA-2225)
Everything _functions_ in our DB migration strategy, but there is room for some refinements.  This PR uses a new dedicated Docker container for DB schema migrations, and streamlines the method for installing the migration utility in the `tests` container.

### Proposed Changes
- When possible, use the Docker image provided by the team that writes our database migration utility
- Install the `golang/migrations` utility in the `tests` container by using `go get` rather than via `apt-get`.

### Security Implications
The new methods of getting the migration utility are at least as trustworthy as what we are currently using.

### Acceptance Validation
Screen Shot 2019-11-04 at 5 33 23 PM

### Feedback Requested
Are there any further steps we should take?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ test:
 	$(MAKE) migrations-test
 
 load-fixtures:
-	docker-compose -f docker-compose.test.yml run --rm -w /go/src/github.com/CMSgov/bcda-ssas-app/db/migrations tests migrate -database "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -path ./ up
+	docker-compose -f docker-compose.test.yml run --rm migrate /migrate -database "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -path db/migrations/ up
 	docker-compose -f docker-compose.yml run ssas sh -c 'tmp/ssas-service --add-fixture-data'
 
 docker-build:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -43,7 +43,7 @@ test:
 	$(MAKE) migrations-test
 
 load-fixtures:
-	docker-compose -f docker-compose.test.yml run --rm -w /go/src/github.com/CMSgov/bcda-ssas-app/db/migrations tests migrate -database "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -path ./ up
+	docker-compose -f docker-compose.test.yml run --rm migrate /migrate -database "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -path db/migrations/ up
```
msnook: I might be overlooking some of the in-depth details.  I wonder if this PR can be simplified to two lines of code in the `Makefile`.  No changes anywhere else:
```
docker pull migrate/migrate
docker run --rm migrate/migrate -database "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -path db/migrations/ up
```

msnook: Might have to mount a volume so the container can see `db/migrations`.  Here is an example where we do that: https://github.com/CMSgov/bcda-ssas-app/blob/master/Makefile#L11

dhgreene: Good advice!  I also had to specify the network.

PR #18, opened 2019-11-19
BCDA-2378 Feature: Allow systems to register IP addresses

dhgreene:
### Fixes [BCDA-2378](https://jira.cms.gov/browse/BCDA-2378)
When ACO's are getting their credentials (registering their systems) through ACO-MS, the process of onboarding should be fully automated.  To make that possible, BCDA needs a piece of information from the ACO that SSAS is not currently providing: what IP address (or addresses) will their system(s) be coming from?

This PR is the first step in getting that piece of information to BCDA: collection and storage.

### Proposed Changes
- Accept an array of IP's during system registration
- Show the IP's
   - In the system registration response
   - In the list of groups (which also shows systems)

How do we fit all the CRUD operations into these minor changes?
* **Create** - system IP's are added during system creation
* **Read** - system IP's are listed with groups
* **Update** - no system update endpoint currently exists, and ACO-MS would not use a dedicated IP update endpoint. If the IP addresses must change, the current system can be deleted and new system can be created.
* **Delete** - IP's are no longer considered valid when the associated system is deleted

### Security Implications
The ACOs' IP addresses are newly stored and transmitted.
In [BCDA-2422](https://jira.cms.gov/browse/BCDA-2422) we will be discussing the larger question of automating IP whitelisting.  The first IP's would start being collected in February.  This PR serves to document the proposed main interface.

### Migration Strategy
This PR requires a new table.  Before deploying this code, the SSAS database schema migration for version 4 should be run from Jenkins.

#### System creation request and response
Screen Shot 2019-11-21 at 9 16 37 AM

#### Group listing showing IP's
Screen Shot 2019-11-21 at 9 17 05 AM

### Feedback Requested
- Should the IP's be stored differently?
- Do the changes to API requests/responses look appropriate, or should we consider alternate naming/structure?
- Do we need further automate testing?

### DIFF
--- a/db/migrations/000004_ips.down.sql
+++ b/db/migrations/000004_ips.down.sql
@@ -1,0 +1,1 @@
+DROP TABLE ips;
--- a/ssas/groups.go
+++ b/ssas/groups.go
@@ -19,11 +19,11 @@ type Group struct {
 }
 
 type SystemSummary struct {
-	ID			uint		`json:"id"`
+	ID        	uint    	`json:"id"`
 	ClientName	string		`json:"client_name"`
-	ClientID	string		`json:"client_id"`
-	IPs			[]string	`json:"ips,omitempty"`
-	UpdatedAt	time.Time	`json:"updated_at"`
+	ClientID	string  	`json:"client_id"`
+	IPs       	[]string	`json:"ips,omitempty"`
+	UpdatedAt 	time.Time	`json:"updated_at"`
 }
 
 type GroupSummary struct {
--- a/ssas/groups.go
+++ b/ssas/groups.go
@@ -22,6 +22,7 @@ type SystemSummary struct {
 	ID         uint      `json:"id"`
 	ClientName string    `json:"client_name"`
 	ClientID   string    `json:"client_id"`
+	IPs		 []string	 `json:"ips,omitempty"`
 	UpdatedAt  time.Time `json:"updated_at"`
 }
 
@@ -108,6 +109,7 @@ func ListGroups(trackingID string) (list GroupList, err error) {
 			ss.ID = system.ID
 			ss.ClientName = system.ClientName
 			ss.ClientID = system.ClientID
+			ss.IPs, _ = system.GetIPs()
 			ss.UpdatedAt = system.UpdatedAt
 
 			gs.Systems = append(gs.Systems, ss)
--- a/ssas/groups.go
+++ b/ssas/groups.go
@@ -19,11 +19,11 @@ type Group struct {
 }
 
 type SystemSummary struct {
-	ID         uint      `json:"id"`
-	ClientName string    `json:"client_name"`
-	ClientID   string    `json:"client_id"`
-	IPs		 []string	 `json:"ips,omitempty"`
-	UpdatedAt  time.Time `json:"updated_at"`
+	ID			uint		`json:"id"`
+	ClientName	string		`json:"client_name"`
+	ClientID	string		`json:"client_id"`
+	IPs			[]string	`json:"ips,omitempty"`
+	UpdatedAt	time.Time	`json:"updated_at"`
 }
 
 type GroupSummary struct {
--- a/ssas/systems.go
+++ b/ssas/systems.go
@@ -357,9 +357,8 @@ func RegisterSystem(clientName string, groupID string, scope string, publicKeyPE
 	}
 
 	for _, address := range ips {
-		_, _, err = net.ParseCIDR(address)
-		if err != nil {
-			regEvent.Help = fmt.Sprintf("invalid CIDR %s; %s", address, err.Error())
+		if !validAddress(address) {
+			regEvent.Help = fmt.Sprintf("invalid IP %s", address)
 			OperationFailed(regEvent)
 			tx.Rollback()
 			return creds, errors.New("error in ip address(es)")
@@ -372,7 +371,7 @@ func RegisterSystem(clientName string, groupID string, scope string, publicKeyPE
 
 		err = tx.Create(&ip).Error
 		if err != nil {
-			regEvent.Help = fmt.Sprintf("could not save CIDR %s; %s", address, err.Error())
+			regEvent.Help = fmt.Sprintf("could not save IP %s; %s", address, err.Error())
 			OperationFailed(regEvent)
 			tx.Rollback()
 			return creds, errors.New("error in ip address(es)")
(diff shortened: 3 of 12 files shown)


db/migrations/000004_ips.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE ips;
```
msnook: Out of curiosity...
Let's say we are at schema version 4 in the DB (i.e., we've already migrated to this version), and we want to perform a migration to bump up to version 5... what happens if we accidentally say input `up 3` instead of `up 5`?  That means we'd be reverting from version 4 to version 3.  In that case, would the version 4 `down` script be executed (dropping the `ips` table)?

ssas/groups.go
```diff
@@ -22,6 +22,7 @@ type SystemSummary struct {
 	ID         uint      `json:"id"`
 	ClientName string    `json:"client_name"`
 	ClientID   string    `json:"client_id"`
+	IPs		 []string	 `json:"ips,omitempty"`
```
msnook: formatting

ssas/systems.go
```diff
@@ -282,14 +290,15 @@ type Credentials struct {
 	ClientSecret string    `json:"client_secret"`
 	SystemID     string    `json:"system_id"`
 	ClientName   string    `json:"client_name"`
+	IPs		   []string	   `json:"ips,omitempty"`
```
msnook: formatting

ssas/systems_test.go
```diff
@@ -428,24 +428,84 @@ func (s *SystemsTestSuite) TestRegisterSystemMissingData() {
...
+		"192.168.0.1/32",			// Single IP's should have a 32-bit mask
+		"2001:db8:a0b:12f0::1/128",	// . . . 128-bit for IPv6
+		"192.168.0.0/24",			// Ranges are good
+		"2001:db8::/32",
+		"10.0.0.0/8",				// We don't care how BIG the range is
+	}
+
+	badIps := []string{
+		"asdf",
+		"192.168.0.1/24",			// Postgres CIDR type does not allow bits to the right of the mask
```
msnook: Nice.

msnook:
Nice.  Does this solution account for the scenario where an ACO has been onboarded (via automated mechanism), but later requires additional IPs to whitelist or later needs to have IPs removed?  Does the interface support updating the list of IPs after an ACO has been onboarded?

db/migrations/000004_ips.down.sql
dhgreene: Yes, that's by design in the migration strategy that there's a rollback option.  You have a good point that this time, a mistake would mean loss of data.

What do you think about tweaking the Jenkins job that applies these changes to the persistent AWS environments?  We could:
- Add a job argument (REVERT?  ROLLBACK?) that would signify your intent to roll back the database version
- Add a step early in the job that queried for the current schema version (the `migrate` utility can do this)
- If the current schema version was higher than the version to migrate to, stop the job unless the ROLLBACK argument was set

dhgreene:
> Nice. Does this solution account for the scenario where an ACO has been onboarded (via automated mechanism), but later requires additional IPs to whitelist or later needs to have IPs removed? Does the interface support updating the list of IPs after an ACO has been onboarded?

We'd need at least one more endpoint to add that capability without having to delete/re-register a system.  I've added a ticket (BCDA-2415) for another endpoint or two that could handle the scenario you're asking about.

db/migrations/000004_ips.down.sql
msnook: Yes, I think we should include that level of safety "just in case".

dhgreene: 👍 I've added BCDA-2416 for that enhancement.

StewGoin:
Does the proposed implementation do any sanity checking on CIDR sizes / validity / 'are we ok with this' type things? I'd hate to accept a 0.0.0.0/0 input from an ACO (or like, their entire corporate /8 or something). Not sure if we've considered that as part of how we automate the human portion of this.

dhgreene:
> Does the proposed implementation do any sanity checking on CIDR sizes / validity / 'are we ok with this' type things? I'd hate to accept a 0.0.0.0/0 input from an ACO (or like, their entire corporate /8 or something). Not sure if we've considered that as part of how we automate the human portion of this.

@StewGoin that should be easy enough to do.  Are we comfortable with an upper limit of `/24` for IPv4 CIDRs?  What size is reasonable for IPv6?

StewGoin:
> > Does the proposed implementation do any sanity checking on CIDR sizes / validity / 'are we ok with this' type things? I'd hate to accept a 0.0.0.0/0 input from an ACO (or like, their entire corporate /8 or something). Not sure if we've considered that as part of how we automate the human portion of this.
> 
> @StewGoin that should be easy enough to do. Are we comfortable with an upper limit of `/24` for IPv4 CIDRs? What size is reasonable for IPv6?

I honestly don't know what's appropriately scoped for the userbase -- my instinct is that a IPv4 /24 is perfectly fine, but I don't know if there are different realities for ACOs. For IPv6 I'd be open to suggestions based on any industry practices we're aware of... it's still a bit of the wild west for range sizes being granted to orgs afaik, but I could be wrong!

dhgreene:
> I honestly don't know what's appropriately scoped for the userbase -- my instinct is that a IPv4 /24 is perfectly fine, but I don't know if there are different realities for ACOs. For IPv6 I'd be open to suggestions based on any industry practices we're aware of... it's still a bit of the wild west for range sizes being granted to orgs afaik, but I could be wrong!

@StewGoin I appreciate your comments here!   We've dropped the CIDR's in favor of individual IP's; if an ACO has a range of 4 addresses they can register the 4 addresses.  Also, look for the [upcoming SIA](https://jira.cms.gov/browse/BCDA-2422) that will address in detail the collection of IP's and whitelisting them in an automated fashion; this code is by way of being an interface for ACO-MS, allowing them to begin planning for the first production IP's that would come in February.

ssas/systems.go
```diff
@@ -595,3 +647,45 @@ func CleanDatabase(group Group) error {
...
 	return nil
 }
+
+func validAddress(address string) bool {
+	ip := net.ParseIP(address)
+	if ip == nil {
+		return false
+	}
+
+	// Source https://en.wikipedia.org/wiki/Reserved_IP_addresses
```
msnook: Awesome!

PR #26, opened 2020-02-10
BCDA-2597 Feature: Log raw group request in SSAS

tbellj:
### Fixes [BCDA-2597](https://jira.cms.gov/browse/BCDA-2597)
To address the extra risk of errors at the integration boundary with ACO-MS, we would like to log the raw requests for managing groups.  This will help troubleshoot any future parsing errors.

### Proposed Changes:
- Add logs for creating and updating groups
- Refactor `ssas.CreateGroup()` to accept a tracking ID as a parameter instead of using the group ID, to support logging before parsing the request

### Security Implications:

All of the data added to the logs in this PR is already present in the logs and database; this change duplicates the logged data in a different format to ensure that it is auditable in the case of a future bug.

##### Creating a group
Screen Shot 2020-02-19 at 11 05 22 PM
##### Editing a group
Screen Shot 2020-02-19 at 11 05 22 PM

### Feedback Requested:
- Much of the refactoring could be avoided if the tracking ID were to be updated after parsing the JSON request body with no errors.  I felt that treating the tracking ID as immutable was a better choice; do you disagree?
- Are there any other suggestions?

### DIFF
--- a/ssas/service/admin/api.go
+++ b/ssas/service/admin/api.go
@@ -34,6 +34,8 @@ import (
 		500: serverError
 */
 func createGroup(w http.ResponseWriter, r *http.Request) {
+	ssas.Logger.Infof("Raw request createGroup: %v", r.Body)
+
 	gd := ssas.GroupData{}
 	err := json.NewDecoder(r.Body).Decode(&gd)
 	if err != nil {
@@ -129,6 +131,7 @@ func listGroups(w http.ResponseWriter, r *http.Request) {
 		500: serverError
 */
 func updateGroup(w http.ResponseWriter, r *http.Request) {
+	ssas.Logger.Infof("Raw request updateGroup: %v", r.Body)
 	id := chi.URLParam(r, "id")
 
 	gd := ssas.GroupData{}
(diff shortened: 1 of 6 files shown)


ssas/service/admin/api.go
```diff
@@ -34,6 +34,8 @@ import (
 		500: serverError
 */
 func createGroup(w http.ResponseWriter, r *http.Request) {
+	ssas.Logger.Infof("Raw request createGroup: %v", r.Body)
```
dhgreene: It would be nice to be able to correlate this log statement with the others related with this API call.  Can we make use of the `ssas.Event` struct (created a few lines later) for that?  In fact, the raw request might even be able to go at the end of the `ssas.Event.Help` field in the `ssas.OperationCalled()` statement.

dhgreene: The resulting format in the logs is a bit rough; running `make test` yields the following line:
```
{"level":"info","msg":"Raw request updateGroup: \u0026{0xc0001bd960 \u003cnil\u003e \u003cnil\u003e false true {0 0} false false false 0x6b5f80}","time":"2020-02-10T15:10:02.755586Z"}
```

I'm thinking the `http.Request.Body` needs some special treatment as an `io.ReadCloser`.

PR #53, opened 2020-10-27
BCDA-3099 - Encode iter count into hash value

mtrang1263:
### Related to [BCDA-3099](https://jira.cms.gov/browse/BCDA-3099)

With [BCDA-3099](https://jira.cms.gov/browse/BCDA-3099), we discovered that our hash computation time is between 100 and 200ms. We want the computation time to take ~1s.

To increase the computation time, we need to increase the iteration count. If we increase the iteration count by increasing `SSAS_HASH_ITERATIONS`, we would invalidate all of current hashes.  To avoid this, we want to encode the iteration count into our hash payload.

Encoding the iteration count is based off of `BKDF2-params` found in [RFC-2898](https://www.ietf.org/rfc/rfc2898.txt)

### Change Details

1. When generating a new hash, we will encode the iteration count as the third parameter.
2. When verifying the hash, we will default to `SSSAS_HASH_ITERATIONS` if we do not have a iteration count inside the hash payload.

### Security Implications

We are storing the iteration count in the hash payload. This is based on [RFC-2898](https://www.ietf.org/rfc/rfc2898.txt)

Security Checklist created: https://confluence.cms.gov/display/BCDA/20-10-28%3A+Encoding+Hash+Iteration+in+Payload+Security+Checklist

### Acceptance Validation
1. Added test to verify backwards compatibility (i.e. no iteration encoded  in the payload).
2. Deployed to dev and validated that iteration count is used when supplied.
3. SSAS postman tests verified that newly created secrets have the iteration count payload present.

### DIFF
--- a/ssas/hash.go
+++ b/ssas/hash.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


msnook:
With the new implementation, what is the computation time?

mtrang1263:
> With the new implementation, what is the computation time?

No new difference in the computation time (iter count is unchanged). This PR is to enable us to update the iteration count without having to invalidate any existing keys

PatKelsh (approved):

PR #56, opened 2020-11-05
BCDA-3847 - Introduce group blacklist capabilities

mtrang1263:
### Fixes [BCDA-3847](https://jira.cms.gov/browse/BCDA-3847)

Adds ability to blacklist a particular group by x_data

#### Background
In our Group model we have a field `x_data`. The field is set by the associated Service Group Administrator. While it is defined as a text field, we expect a JSON encoded payload.

Since it's defined by the SGA and does not conform to any pre-defined schema, the `x_data` field was not used with SSAS. We currently embedded the CMS_ID associated with the group for use in the bcda-app.

#### Implementation
Since we assume that the `x_data` is JSON encoded string, we had considered implementing exact string matching for the associated group blacklist. So if a group had the following: `x_data:{"cms_ids":["A1234"]}` and ACO A1234 from our system, we would insert a string field of `{"cms_ids":["A1234"]}` or `x_data:{"cms_ids":["A1234"]}`. 

However, this approach does not leverage the fact that x_data contains a set of key:value pairs. Instead of using exact string matching, we decided to use a key:value blacklist. Using key:value blacklist, we can restrict access to ACO A1234 by adding the following entry: `{"cms_ids":"A1234"}`. After adding this record, any attempt to onboard a group who has an x_data field containing `cms_ids:"A1234"` will not be inserted. If a caller sent the following: `x_data:{"cms_ids" : [ "A1234" ]}` (notice the spaces), we would still consider this group to be blacklisted. If we had used exact string matching, we would not have a match and the group would not be blacklisted.

By using the JSON structure, we allow flexibility with the data we can store in the x_data field without having to regenerate any new group blacklist entries.

The implementation of this strategy can be found in blacklist.go#checkXData

### Change Details
1. Refactored the blacklist implementation to allow us to extend the implementation to multiple types. It used to implement a token blacklist, but now it has been extended to implement group blacklists.
2. Introduced blacklist.GroupEntry type.
3. Added schema change that'll onboard the group blacklist schema.

### Security Implications

We will be storing information that will allow us to blacklist groups. This is needed for ACO termination.

### Acceptance Validation
1. CI to test functionality.

Upcoming tickets will be implemented to use this functionality. Once we have those in place, we can introduce smoke tests.


ssas/service/blacklist/blacklist.go
```diff
@@ -0,0 +1,325 @@
...
+
+func (b *blacklist) IsGroupBlacklisted(group ssas.Group) bool {
+	var cacheItems map[string]cache.Item
+
+	b.RLock()
+	cacheItems = b.gc.Items()
+	b.RUnlock()
+
+	var xDataObj map[string]interface{}
+	if err := json.Unmarshal([]byte(group.XData), &xDataObj); err != nil {
```
dhgreene, line 158: Historically we've tried to keep SSAS unaware of the contents of `group.XData`.  Yes, we are expecting it to be in JSON format, but we're accepting text and so far the only manipulation we're doing is `strconv.Unquote()`.  For the purposes of the blacklist, what if we:

- Looked at XData as a string, not JSON
- Ignored the key value we're expecting (`cmsids`)
- Just searched to see whether the group ID (OrgID) in the group blacklist was present in the string

The benefit would be that SSAS continues knowing as little as possible about this data packet being passed along.

ssas/service/blacklist/blacklist.go
```diff
@@ -0,0 +1,325 @@
...
+func Start() {
+	// Guarantee that we only have a single Blacklist instance running
+	if Blacklist != nil {
+		return
+	}
+	Blacklist = newBlacklist(24*time.Hour,
+		time.Duration(cfg.GetEnvInt("SSAS_TOKEN_BLACKLIST_CACHE_CLEANUP_MINUTES", 15))*time.Minute,
+		time.Duration(cfg.GetEnvInt("SSAS_GROUP_BLACKLIST_CACHE_CLEANUP_MINUTES", 15))*time.Minute,
+		time.Duration(cfg.GetEnvInt("SSAS_TOKEN_BLACKLIST_CACHE_TIMEOUT_MINUTES", 60*24))*time.Minute,
+		time.Duration(cfg.GetEnvInt("SSAS_GROUP_BLACKLIST_CACHE_TIMEOUT_MINUTES", 60*24))*time.Minute,
```
dhgreene, line 31: I may not be understanding the use case, but I'm surprised that we're only blacklisting an org for 24 hours.

ssas/service/blacklist/blacklist.go
```diff
@@ -0,0 +1,325 @@
...
+	// Need to acquire a lock since we're clearing the entire cache.
+	// Any reads in-between us re-hydrating the cache is invalid (false negatives)
+	b.Lock()
+	defer b.Unlock()
+
+	b.tc.Flush()
+
+	// If the key already exists in the cache, it will be updated.
+	for _, entry := range entries {
+		cacheDuration := time.Until(time.Unix(0, entry.CacheExpiration))
```
mtrang1263, line 230: This used to be the following:
```
cacheDuration := time.Since(time.Unix(0, entry.CacheExpiration))
```

This resolves to  `time.Now().Sub(time.Unix(0, entry.CacheExpiration))`. If the cache is not expired, we expect a time value that is greater than time.Now(). Using that operation, it would result in a negative time.Duration value. This causes the cache entry to never expire (See https://github.com/patrickmn/go-cache/blob/master/cache.go#L51 for more details).

This does not poise a large issue since we are completely rebuilding the cache, but needs to be resolved in case we no longer rebuild the cache.

Reference to the [old impl](https://github.com/CMSgov/bcda-ssas-app/pull/56/files#diff-020d650a2ce74e27bb6e3de0c7d301c9c192a5c033981502f474e570a695e79eL110)

ssas/service/blacklist/blacklist.go
```diff
@@ -0,0 +1,325 @@
...
+	if entries, err = getUnexpiredTokenEntries(); err != nil {
+		return err
+	}
+
+	// Need to acquire a lock since we're clearing the entire cache.
+	// Any reads in-between us re-hydrating the cache is invalid (false negatives)
+	b.Lock()
+	defer b.Unlock()
+
+	b.tc.Flush()
```
mtrang1263, line 226: The [old implementation ](https://github.com/CMSgov/bcda-ssas-app/pull/56/files#diff-020d650a2ce74e27bb6e3de0c7d301c9c192a5c033981502f474e570a695e79eL106) did not use a mutex to restrict access to the cache while it was being rebuilt. This may result in false negatives being reported (token is blacklisted, but we **do not** find it in the cache).

If a caller checks to see if the token is blacklisted after we've flushed (cleared) the cache and before we've finished rebuilding the cache, we would incorrectly say that the token **is not** blacklisted.

By acquiring a write lock, we prevent callers from reading the cache until it's completely rebuilt.

ssas/service/blacklist/blacklist.go
```diff
@@ -0,0 +1,325 @@
...
+	ssas.ServiceStarted(event)
+
+	ticker = time.NewTicker(refreshFreq)
+	done = make(chan struct{})
+
+	go func() {
+		for {
+			select {
+			case <-done:
+				return
```
mtrang1263, line 199: To avoid leaking the go routine after we've stopped the ticker, we provide a signal channel that we'll use to tell the goroutine to exit. See [here](https://github.com/CMSgov/bcda-ssas-app/pull/56/files#diff-c641dc75064eafa5c9761e8a218e274f9984fd0b8e000ab3d5b8e473677e851aR264) for its usage.

ssas/service/blacklist/token.go
```diff
@@ -0,0 +1,73 @@
+package blacklist
```
mtrang1263, line 1: Replaces ssas/blacklist.go

ssas/service/blacklist/blacklist.go
mtrang1263: Good call - I think we'll look at requiring the fields to be defined by the application (and not provide a default value).

ssas/service/blacklist/blacklist.go
mtrang1263: > Historically we've tried to keep SSAS unaware of the contents of `group.XData`. Yes, we are expecting it to be in JSON format

Definitely agreed that we should continue keeping the XData field opaque to SSAS. The only extent that we're inspecting the field is to verify that the field can be deserialized into a JSON object. The actual logic checks can be found in the `checkXData` function. The implementation does not make any assumptions on what is contained in the JSON object. Instead walks through the tree and attempts to find a matching entry in our blacklist tree.

> * Looked at XData as a string, not JSON
> * Ignored the key value we're expecting (cmsids)

Might not be fully understanding, but I'm not sure how we would know what key values are present and should be ignored. If we put in a list of those keys (like cmsids), I think we're actually adding more knowledge of what is going into the xdata.

> Just searched to see whether the group ID (OrgID) in the group blacklist was present in the string

Definitely! Our schema designates a the Group field to check against. For the near short term, we want to support xdata checks. We should look at expanding it to other fields if we there's a need.

dhgreene: Yeah, that wasn't clear.  Here's another way to say it: SSAS doesn't care what's in XData.  It shouldn't care what format it's in either.  A mostly-data-blind blacklist can be implemented by blacklisting a string that _shall not_ be present in XDATA--in this case, the string would be an org id.

dhgreene: The format of XData is a contract between the other parties using SSAS.

mtrang1263:
Speaking a bit more with @msnook and @dhgreene , there doesn't seem to be much benefit of adding an extensible blacklist to SSAS right now. We may want to consider/re-open this PR if and when other services start leveraging SSAS and have a need for a generic group blacklist capability.

As such, we'll close this PR and implement the ACO blacklist functionality on the bcda-app side.

PR #57, opened 2020-11-05
BCDA-3847 - Resolve issues found in tokenblacklist impl

mtrang1263:
### Discovered when working on [BCDA-3847](https://jira.cms.gov/browse/BCDA-3847)

While working on the extension of the blacklist, we discovered some possible bugs in tokenblacklist.go.

1. Race condition when the token cache is rebuilt. We can encounter false negatives (token should be blacklisted, but we consider it to not be) when we are rebuilding the cache from the database.
2. Resource leak when stopping the cache refresh ticker. We never finish the associated goroutine.
3. We were computing a negative cache lifetime when loading token entries from the database. This does not have much real world impact since the cache is routine rebuilt.

### Change Details

1. Use a RWMutex to prevent reads from occurring when we are rebuilding the cache.
2. Return a CancelFunc to allow us to terminate the goroutine that was created when starting the cache refresh ticker.
3. Use `time.Until()` to compute the remaining time the token should exist in the cache.

### Security Implications

### Acceptance Validation

CI Passes

### DIFF
--- a/ssas/service/tokenblacklist.go
+++ b/ssas/service/tokenblacklist.go
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	TokenBlacklist Blacklist
+	TokenBlacklist *Blacklist
 	// This default cache timeout value should never be used, since individual cache elements have their own timeouts
 	defaultCacheTimeout  = 24 * time.Hour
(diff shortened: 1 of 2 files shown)


msnook (approved):
LGTM!

PR #60, opened 2020-12-03
Sl/bcda 3921 create cli to identify aco

SMLuthi:
### Fixes [BCDA-3921](https://jira.cms.gov/browse/BCDA-3921)

Due to an incident it was determined that we needed a way to identify ACO's by either supplying their client_id or their API key.

### Proposed Changes

Create a CLI command that takes in either the client id or the api key and prints out the ACO's xdata field from `Group`. This will supply the ACO's ID.

### Change Details

Just as noted above. The main.go now has a new cli flag `get-xdata` that requires either the `client-id` or the `api-key` flag to be set as well. Either of these identifiers will allow the cli tool to gather the xdata for the ACO and print it to stdout.

Note: If both identifiers are provided, the API key will supersede the client id.
 
### Security Implications

This is slightly difficult for me to answer as I dont know the eco system that well yet. I'm not sure what else this xdata field holds but, regardless, it will dumped to stdout and logged. These logs are most likely accessible somewhere else (ie Splunk?). However, this is all being run behind our firewall so if someone could access to that then running this cli command or retrieving the results are probably the least of our concerns.

### Acceptance Validation

AC has been met. We can now retrieve the ACO id for a given secret.

### Feedback Requested

As always, let me know if there is a more GOmatic way to do things! Also, let me know if I misinterpreted anything with my naming... and should rename something to be more accurate. The words ACO, system, client, etc... are thrown around a lot and I want to make sure I used them accurately.

Also if there are some other tests cases that would be valuable.

### DIFF
--- a/ssas/service/main/main.go
+++ b/ssas/service/main/main.go
@@ -30,6 +30,7 @@ import (
 	"database/sql"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -54,10 +55,10 @@ var doResetSecret bool
 var doNewAdminSystem bool
 var doListIPs bool
 var doListExpCreds bool
-var doGetXdata bool
+var doShowXData bool
 var doStart bool
 var clientID string
-var APIKey string
+var auth string
 var systemName string
 var output io.Writer
 
@@ -80,9 +81,9 @@ func init() {
 	const usageListExpCreds = "list credentials about to expire or timeout due to inactivity"
 	flag.BoolVar(&doListExpCreds, "list-exp-creds", false, usageListExpCreds)
 
-	const usageGetXdata = "list ACOs group xdata"
-	flag.BoolVar(&doGetXdata, "get-xdata", false, usageGetXdata)
-	flag.StringVar(&APIKey, "api-key", "", "a systems api key")
+	const usageShowXData = "display ACOs group xdata"
+	flag.BoolVar(&doShowXData, "show-xdata", false, usageShowXData)
+	flag.StringVar(&auth, "auth", "", "an auth header containing the hashed client id")
 
 	const usageStart = "start the service"
 	flag.BoolVar(&doStart, "start", false, usageStart)
@@ -116,8 +117,15 @@ func main() {
 		listExpiringCredentials()
 		return
 	}
-	if doGetXdata && (clientID != "" || APIKey != "") {
-		getXData(clientID, APIKey)
+	if doShowXData && (clientID != "" || auth != "") {
+		if clientID != "" || auth != "" {
+			err := showXData(clientID, auth)
+			if err != nil {
+				ssas.Logger.Error(err)
+			}
+		} else {
+			ssas.Logger.Error("`show-xdata` requires either the client-id or auth key arg be set")
+		}
 		return
 	}
 	if doStart {
@@ -351,23 +359,24 @@ func cliTrackingID() string {
 	return fmt.Sprintf("cli-command-%d", time.Now().Unix())
 }
 
-func getXData(clientID, APIkey string) {
-	if APIkey != "" {
+func showXData(clientID, auth string) error {
+	// The auth header decoding logic was pulled from Go's requuest.go#parseBasicAuth func
+	if auth != "" {
 		const prefix = "Basic "
-		// Check that the APIkey provided that has more than just the `Basic` prefix
-		if len(APIkey) < len(prefix) || !strings.EqualFold(APIkey[:len(prefix)], prefix) {
-			panic("must provide a valid API key")
+		// Check that the auth provided that has more than just the `Basic` prefix
+		if len(auth) < len(prefix) || !strings.EqualFold(auth[:len(prefix)], prefix) {
+			return errors.New("must provide a valid auth hash")
 		}
-		c, err := base64.StdEncoding.DecodeString(APIkey[len(prefix):])
+		c, err := base64.StdEncoding.DecodeString(auth[len(prefix):])
 		if err != nil {
-			panic("unable to decode the API key: " + err.Error())
+			return errors.New("unable to decode the auth hash: " + err.Error())
 		}
 
 		cs := string(c)
 		// Get length of string up to colon in string (ie the client id length)
 		s := strings.IndexByte(cs, ':')
 		if s < 0 {
-			panic("No client id present after decoding API key")
+			return errors.New("no client id present after decoding auth hash")
 		}
 
 		clientID = cs[:s]
@@ -375,13 +384,15 @@ func getXData(clientID, APIkey string) {
 
 	system, err := ssas.GetSystemByClientID(clientID)
 	if err != nil {
-		panic("invalid client id: " + err.Error())
+		return errors.New("invalid client id: " + err.Error())
 	}
 
 	group, err := ssas.GetGroupByGroupID(system.GroupID)
 	if err != nil {
-		panic("unable to find group with id " + system.GroupID + ": " + err.Error())
+		return errors.New("unable to find group with id " + system.GroupID + ": " + err.Error())
 	}
 
 	fmt.Fprintln(output, group.XData)
+
+	return nil
 }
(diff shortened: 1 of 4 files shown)


ssas/service/main/main.go
```diff
@@ -107,6 +116,10 @@ func main() {
 		listExpiringCredentials()
 		return
 	}
+	if doGetXdata && (clientID != "" || APIKey != "") {
+		getXData(clientID, APIKey)
```
mtrang1263, line 120: Would it make sense to return an error if the caller does not supply the clientID or APIKey?

ssas/service/main/main.go
```diff
@@ -337,3 +350,38 @@ func closeRows(rows *sql.Rows) {
 func cliTrackingID() string {
 	return fmt.Sprintf("cli-command-%d", time.Now().Unix())
 }
+
+func getXData(clientID, APIkey string) {
```
mtrang1263, line 354: since the name is `getXData`, would it make sense for us to return the XData value and display it outside of the function?

ssas/service/main/main.go
```diff
@@ -337,3 +350,38 @@ func closeRows(rows *sql.Rows) {
...
+	if APIkey != "" {
+		const prefix = "Basic "
+		// Check that the APIkey provided that has more than just the `Basic` prefix
+		if len(APIkey) < len(prefix) || !strings.EqualFold(APIkey[:len(prefix)], prefix) {
+			panic("must provide a valid API key")
```
mtrang1263, line 359: Instead of panicking here (and the other spots), can we return an error instead? That way the caller can determine what it wants to do (log the information and then return an exit code?)

ssas/service/main/main.go
```diff
@@ -337,3 +350,38 @@ func closeRows(rows *sql.Rows) {
 func cliTrackingID() string {
 	return fmt.Sprintf("cli-command-%d", time.Now().Unix())
 }
+
+func getXData(clientID, APIkey string) {
+	if APIkey != "" {
+		const prefix = "Basic "
```
mtrang1263, line 356: Does it make sense to call out that this implementation is lifted from Go's request.go#parseBasicAuth?

ssas/service/main/main_test.go
```diff
@@ -28,6 +33,43 @@ func (s *MainTestSuite) TestResetSecret() {
...
+	creds, _ := s.createACOData()
+
+	// Build encoded api key to mimic auth header
+	APIKey := "Basic " + base64.StdEncoding.EncodeToString([]byte(creds.ClientID+":"+creds.ClientSecret))
+
+	output := captureOutput(func() { getXData("", APIKey) })
+	assert.Equal(s.T(), "fake x_data\n", output)
+}
+
+func (s *MainTestSuite) createACOData() (creds ssas.Credentials, group ssas.Group) {
```
mtrang1263, line 53: Would it make sense to move `createACOData()` (and `createIntrospectData()`) into the testutils.go file?

We can update the signature to accept the *testing.T object so you can still use the same assertions.

ssas/service/main/main.go
SMLuthi: The reason I stuck with using `panic` instead of bubbling up an error is mostly for keeping it consistent with house the other cli commands are built. But I can revision to make it to do that if we find more value in that.

mtrang1263: I think it's good for us to start return errors instead. In the case of the CLI, we can offer richer error messaging/formatting. And in a broader sense, returning errors allows the caller of the function to use some alternative strategy (retries, log, etc.) which would be more difficult to reason about if we were using panics

ssas/service/main/main.go
DeirdreHolub: Why would something called APIkey be prefixed with "Basic "? This suggests that what you're actually expecting to be passed in is the value string used in a basic auth header. Why would we pass that whole thing in to a CLI command? why not just pass in the base 64 encoded token by itself?

msnook (changes requested):
To be honest, I was expecting to see this CLI in `bcda-app`.  If we include it in `bcda-ssas-app`, then we've coupled SSAS with ACOs, which goes against the design.  Is there a reason for this being in SSAS instead of ACO API?

ssas/service/main/main.go
SMLuthi: Very true, i'll rename this to `auth` as it seems thats what Go likes to call it (and it makes sense)

SMLuthi:
@msnook Martin and I actually looked at this originally and opted to to include it in `bcda-ssas-app`. I dont think this actually couples SSAS with ACOs as all it is doing is retrieving a specific field (`xdata`) and displaying that in the terminal. Its not doing any actual work with that data nor does it care what it contains. The main reason this was put in here, though, was because `bcda-app` doesn't have the database context to access this data in the first place (as I understand it). So it would be, as I understand it, impossible for `bcda-app` to access.

msnook:
> @msnook Martin and I actually looked at this originally and opted to to include it in `bcda-ssas-app`. I dont think this actually couples SSAS with ACOs as all it is doing is retrieving a specific field (`xdata`) and displaying that in the terminal. Its not doing any actual work with that data nor does it care what it contains. The main reason this was put in here, though, was because `bcda-app` doesn't have the database context to access this data in the first place (as I understand it). So it would be, as I understand it, impossible for `bcda-app` to access.

Ahh, so a more general utility which simply retrieves `xdata`.  Makes sense!  I wonder, then, if we should remove all references to the term `ACO` that was introduced in this PR?  For example, see [here](https://github.com/CMSgov/bcda-ssas-app/pull/60/files#diff-ed9f8507fc284a4bca464b96c1d9b0f1ab50c89fe7413d1db7dcc0cea44926f7R83) (and other places).  Can it be more general, removing the concept of `ACO` from SSAS?

SMLuthi:
@msnook Absolutely! Im up for a recommendation, as I dont know the nomenclature very well. What would be more appropriate?

ssas/service/main/main.go
```diff
@@ -78,8 +81,15 @@ func init() {
 	const usageListExpCreds = "list credentials about to expire or timeout due to inactivity"
 	flag.BoolVar(&doListExpCreds, "list-exp-creds", false, usageListExpCreds)
 
+	const usageShowXData = "display ACOs group xdata"
```
msnook, line 84: To eliminate usage of ACO, maybe change to `display group xdata` to be more general.

ssas/service/main/main_test.go
```diff
@@ -28,6 +32,29 @@ func (s *MainTestSuite) TestResetSecret() {
 	assert.NotEqual(s.T(), "", output)
 }
 
+func (s *MainTestSuite) TestShowXDataWithClientID() {
+	creds, _ := ssas.CreateACOData(s.T(), s.db)
```
msnook, line 36: Eliminate usage of `ACO` to be more general.  Maybe just `CreateTestXData()`?

ssas/testutils.go
```diff
@@ -110,3 +116,23 @@ func someRandomBytes(n int) ([]byte, error) {
 	}
 	return b, nil
 }
+
+func CreateACOData(t *testing.T, db *gorm.DB) (creds Credentials, group Group) {
```
msnook, line 120: `CreateTestXData()`?

msnook:
> @msnook Absolutely! Im up for a recommendation, as I dont know the nomenclature very well. What would be more appropriate?

I added suggestions as comments in the code.

msnook (approved):

ssas/service/main/main.go
```diff
@@ -337,3 +358,36 @@ func closeRows(rows *sql.Rows) {
 func cliTrackingID() string {
 	return fmt.Sprintf("cli-command-%d", time.Now().Unix())
 }
+
+func showXData(clientID, auth string) error {
+	// The auth header decoding logic was pulled from Go's requuest.go#parseBasicAuth func
+	if auth != "" {
+		c, err := base64.StdEncoding.DecodeString(auth)
+		if err != nil {
+			return errors.New("unable to decode the auth hash: " + err.Error())
```
mtrang1263, line 367: You can also use `fmt.Errorf` so you don't have to do all of this string concatenation.

So this could be:
```
return fmt.Errorf("unable to decode the auth hash: %w", err)
```

The `%w` allows you to leverage Go's error wrapping/chaining

Ref: https://blog.golang.org/go1.13-errors

mtrang1263 (dismissed):
LGTM! Just put a comment on alternative ways to generate the `error` struct.

mtrang1263 (approved):

PR #61, opened 2020-12-10
BCDA-3917 Determine SSAS_HASH_ITERATIONS value

adamayd:
### Fixes [BCDA-3917](https://jira.cms.gov/browse/BCDA-3917)

We want our hash generation in SSAS to take ~1s.

We had discovered that not all entries in the secrets table have the iteration count encoded. This will prevent us from updating the env var w/o invalidating the credentials associated with those hash entries. 

### Proposed Changes

Create a stand alone test that does not run in CI/CD that allows to test the number of iterations needed to reach 1s.

### Change Details

Added TestHashIterTime test that uses a Skip conditional based on a local / runtime specified environment variable.

### Security Implications

This test itself has no security implications.

### Acceptance Validation

We have determined the hash iteration count that will take ~1s to generate.

### Feedback Requested

Go best practices, different approach to solution.

### DIFF
--- a/ssas/hash_test.go
+++ b/ssas/hash_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/pborman/uuid"
 	"github.com/stretchr/testify/assert"
@@ -63,6 +64,25 @@ func (s *HashTestSuite) TestHashWithIter() {
 
 	s.True(h.IsHashOf(val))
 	s.True(h1.IsHashOf(val))
+}
+
+func (s *HashTestSuite) TestHashIterTime() {
+	var resultArr [5]int64
+	hashIter = 130000
+	hashKeyLen = 64
+	saltSize = 32
+	for i := 0; i < 5; i++ {
+		start := time.Now()
+
+		val := uuid.New()
+		_, err := NewHash(val)
+		s.NoError(err)
+
+		elapsed := time.Now().Sub(start).Milliseconds()
+		resultArr[i] = elapsed
+		fmt.Println(elapsed)
+	}
+	// Print average of results
 }
 
 func TestHashTestSuite(t *testing.T) {
--- a/ssas/hash_test.go
+++ b/ssas/hash_test.go
@@ -2,6 +2,7 @@ package ssas
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


ssas/hash_test.go
```diff
@@ -65,6 +67,35 @@ func (s *HashTestSuite) TestHashWithIter() {
...
+	// This test is not intended to run with the remain suite.  Is is solely used to determine the number of hash
+	// iterations needed to reach a 1 second duration. To invoke the test use:
+	// DEBUG=true  go test -v github.com/CMSgov/bcda-ssas-app/ssas -run TestHashIterTime
+
+	// Check for local dev environment
+	if os.Getenv("DEBUG") != "true" {
+		t.SkipNow()
+	}
+
+	var resultTime int64
```
mtrang1263, line 80: Nit - You can actually keep the resultTime as a time.Duration value (since [time.Duration is also an int64](https://golang.org/pkg/time/#Duration)).

If you keep resultTime as a time.Duration value, you can leverage [time.Duration's String() function](https://golang.org/pkg/time/#Duration.String) which will pretty print the units for you. That way you don't have to assume ms units.

If you decide on that approach, you'll want to cast runCount to a time.Duration value when you are computing the average.

ssas/hash_test.go
```diff
@@ -65,6 +67,35 @@ func (s *HashTestSuite) TestHashWithIter() {
 	s.True(h1.IsHashOf(val))
 }
 
+func TestHashIterTime(t *testing.T) {
+	// This test is not intended to run with the remain suite.  Is is solely used to determine the number of hash
+	// iterations needed to reach a 1 second duration. To invoke the test use:
+	// DEBUG=true  go test -v github.com/CMSgov/bcda-ssas-app/ssas -run TestHashIterTime
```
mtrang1263, line 73: Really nice description!

ssas/hash_test.go
```diff
@@ -65,6 +67,35 @@ func (s *HashTestSuite) TestHashWithIter() {
...
+
+	// Check for local dev environment
```
mtrang1263, line 75: Nit - I don't think the comment here is adding much value. The if statement seems to convey enough

ssas/hash_test.go
```diff
@@ -65,6 +67,35 @@ func (s *HashTestSuite) TestHashWithIter() {
...
+	if os.Getenv("DEBUG") != "true" {
+		t.SkipNow()
```
mtrang1263, line 77: :+1: Very elegant way to selectively run our tests. By using skipNow(), we won't be running this test in our unit tests, but it's easy for users to quickly enable for local testing.

ssas/hash_test.go
mtrang1263, line 80: Nit 2 -Instead of resultTime, what do you think about totalTime? Since we're accumulating the various resultTimes?

adamayd: Updated resultTime(totalTime) to be defined as a time.Duration type.

ssas/hash_test.go
adamayd: Removed

ssas/hash_test.go
adamayd: Renamed to totalTime

mtrang1263 (approved):
LGTM!

PR #62, opened 2020-12-16
BCDA-3941 - Update gorm dependency to v2

mtrang1263:
### Fixes [BCDA-3941](https://jira.cms.gov/browse/BCDA-3941)

We are using an outdated version of gorm (v1). We want to move to the latest version (v2)

### Change Details
1. Updating gorm dependency from v1 to v2.
2. Resolving breaking changes between v1 and v2 API.
3. Rework queries/models based on the v1 to v2 changes.

Used [work](https://github.com/CMSgov/bcda-app/pull/601) from the bcda-app migration to help with breaking changes.

### Security Implications
Upgrading from gorm v1 to gorm v2 

### Acceptance Validation
1. CI passes.
2. Deploy to dev environment to verify everything works as expected.


ssas/groups.go
```diff
@@ -23,7 +24,7 @@ type SystemSummary struct {
 	GID        uint      `json:"-"`
 	ClientName string    `json:"client_name"`
 	ClientID   string    `json:"client_id"`
-	IPs        []string  `json:"ips,omitempty"`
+	IPs        []string  `json:"ips,omitempty" gorm:"-"`
```
mtrang1263, line 27: Had to change because gorm was returning an error around not being able to handle the IPs field since it's not defined in the Systems table.

We can tie a foreign key between System and IPs tables, but leaving it out for now since it was not being done with gorm v1.

ssas/systems.go
```diff
@@ -45,23 +44,6 @@ func getEnvVars() {
 	CredentialExpiration = time.Duration(expirationDays*24) * time.Hour
 }
 
-/*
-	InitializeSystemModels will call gorm.DB.AutoMigrate() for models associated with systems, and set up foreign key
-	relationships for those models if needed
-*/
-func InitializeSystemModels() *gorm.DB {
```
mtrang1263, line 52: No longer needed since we've migrated to golang migrate https://github.com/golang-migrate/migrate

ssas/groups.go
```diff
@@ -199,7 +200,7 @@ func cascadeDeleteGroup(group Group) error {
 	)
 	defer Close(db)
 
-	err := db.Table("systems").Where("group_id = ?", group.GroupID).Pluck("ID", &systemIds).Error
+	err := db.Table("systems").Where("group_id = ?", group.GroupID).Pluck("id", &systemIds).Error
```
SMLuthi, line 203: v2 is now case sensitive?

SMLuthi (approved):
LGTM!

adamayd (approved):
LGTM

ssas/groups.go
mtrang1263: Ah, it's not. I think this change was part of me troubleshooting some other errors. I'l leave it as is though if that's OK with you

PR #70, opened 2021-04-27
BCDA 4452 Fix: Pin Postgres image version to 10 

lostpurplecow:
### Fixes [BCDA-4452](https://jira.cms.gov/browse/BCDA-4452)

The current deployed major version of Postgres in AWS for ACO API is Postgres 10.  However, in the docker-compose stacks for bcda-ssas-app, we are pulling from postgres:latest (see here for example).

We need to ensure all references to pulling Postgres images from Dockerhub bcda-ssas-app align with the version of Postgres that we are using in AWS.  This should be `postgres:10` image in Dockerhub.

### Proposed Changes

- Pin `postgres:10` version

### Security Implications

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3'
 
 services:
   db:
-    image: postgres
+    image: postgres:10
     environment:
       - POSTGRES_DB=bcda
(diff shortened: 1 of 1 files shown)


CosmicSyntax (approved):
LGTM

PR #77, opened 2021-06-03
Sg/dpc 1380 v2 token info

MrMorie:
### Fixes [DPC-1380](https://jira.cms.gov/browse/DPC-1380)

Much like the SSAS v1 POST /introspect endpoint, the purpose of POST /v2/token_info is so that DPC or the ACO API can find out whether an access token is valid.  Unlike SSAS v1 POST /introspect:

This endpoint will be in the admin service
This endpoint will return the parsed token claims–most especially the contents of the XData element
### Initial Proposed Changes
Add /token_info endpoint to Admin SSAS

### Change Details
Facts:
- SSAS Admin and Public services have different public keys (specified as env variables)
- Adding the token_info API to the Admin service would require one of the following:
- - Use the same key pair for admin and public services, or
- - Introduce a new variable to the admin service with the location of the public service's key (would require more work)

In order to avoid sharing keys between public and private services it was agreed that we would move the token_info API to the public service. SSAS is an internal service and DPC API would simply not proxy any external requests to token_info.

### Security Implications

### Acceptance Validation

Screen Shot 2021-06-08 at 12 30 26 PM
Screen Shot 2021-06-08 at 12 29 54 PM


ssas/service/public/api.go
```diff
@@ -720,3 +720,39 @@ func introspect(w http.ResponseWriter, r *http.Request) {
...
+		response["valid"] = false
+	} else {
+		claims := jwt.MapClaims{}
+		if _, _, err := new(jwt.Parser).ParseUnverified(tokenS, claims); err != nil {
+			ssas.Logger.Infof("could not unmarshal access token")
+			jsonError(w, http.StatusText(http.StatusInternalServerError), "internal server error")
+			return
+		}
+		response["valid"] = true
+		response["claims"] = claims
```
dhgreene, line 754: The output here is tightly coupled with the JWT we're creating. Really the only information we need to respond with is the "dat" claim.  What if we did this instead?
```suggestion
		response["data"] = claims["dat"]
```

ssas/service/public/api_test.go
```diff
@@ -906,3 +906,100 @@ func mintClientAssertion(issuer string, subject string, aud string, issuedAt int
...
+	rCtx := chi.NewRouteContext()
+	req = req.WithContext(context.WithValue(req.Context(), chi.RouteCtxKey, rCtx))
+	handler := http.HandlerFunc(validateAndParseToken)
+	rr := httptest.NewRecorder()
+	handler.ServeHTTP(rr, req)
+	assert.Equal(s.T(), http.StatusOK, rr.Result().StatusCode)
+	assert.Contains(s.T(), rr.Body.String(), `{"valid":false}`)
+}
+
+func (s *APITestSuite) MintTestAccessTokenWithDuration(duration time.Duration) (*jwt.Token, string, error) {
```
dhgreene, line 987: It's worth trying to get a token in a different way than adding more token-creation code. This might be a good place to take advantage of (and update?) [this](https://github.com/CMSgov/bcda-ssas-app/blob/master/ssas/service/test_utils.go).

ssas/service/public/api.go
MrMorie: Done.

ssas/service/public/api_test.go
MrMorie: We can grab an access token by calling the /token endpoint but we still need this method to create very short lived tokens.

We could modify the servers token duration, but that may introduce race conditions if tests are run in parallel.

MrMorie:
Build & Package passed https://bcda-ci.adhocteam.us/view/SSAS/job/SSAS%20-%20Build%20and%20Package/1424/

MrMorie:
BCDA Dev deploy passed https://bcda-ci.adhocteam.us/job/BCDA%20-%20Build%20and%20Package/2534/

dhgreene (approved):
🚀

PR #82, opened 2021-06-14
DPC-1444 generate macaroon as client token

MrBilnon:
### Fixes [DPC-1444](https://jira.cms.gov/browse/DPC-1444)

Need to return a macaroon for the client token in v2 system creation. 

### Proposed Changes

Create a RootKey entity that stores the secret that is used to generate the macaroon with the expiration date and the id that will be used in the ClientToken entity as a foreign key to RootKey.

### Change Details

* Edit SaveClientToken to generate a RootKey and the associated macaroon and return it from save client token to be used as the client token in the credentials

### Security Implications

### Acceptance Validation

Screen Shot 2021-06-15 at 1 58 00 PM
Screen Shot 2021-06-15 at 1 58 17 PM


ssas/root_key.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/pborman/uuid"
+	"gopkg.in/macaroon.v2"
+	"gorm.io/gorm"
+	"math/big"
+	"time"
+)
+
+type RootKey struct {
+	gorm.Model
+	Uuid      string
```
jonfulk, line 17: I'm surprised `gofmt` doesn't want this to be `UUID`

ssas/systems.go
```diff
@@ -74,31 +76,44 @@ type Secret struct {
...
+		ExpiresAt: rk.ExpiresAt,
 	}
 
 	if err := db.Create(&ct).Error; err != nil {
-		return fmt.Errorf("could not save client token for clientID %s: %s", system.ClientID, err.Error())
+		return "", fmt.Errorf("could not save client token for clientID %s: %s", system.ClientID, err.Error())
 	}
 	ClientTokenCreated(Event{Op: "SaveClientToken", TrackingID: uuid.NewRandom().String(), ClientID: system.ClientID})
-	return nil
+	return token, nil
```
jonfulk, line 116: is `token` just the stringified version of `ct`? I'm curious why you save `ct` but return `token`

MrBilnon: token is what get's generated by the macaroon library, we don't save the macaroon in the database only the uuid tied to the macaroon and the secret used to generate the macaroon.  This is similar to how dpc auth did it, also you can see the original code never returned ct and was returning `ct := b64.StdEncoding.EncodeToString([]byte("client-token-placeholder"))` further down in this file

ssas/systems.go
```diff
@@ -74,31 +79,44 @@ type Secret struct {
...
+	SystemID  uint      `json:"system_id"`
+	ExpiresAt time.Time `json:"expires_at"`
 }
 
 /*
 	SaveClientToken should be provided with a token label and toke uuid, which will
 	be saved to the client tokens table and associated with the current system.
 */
-func (system *System) SaveClientToken(uuidS string, label string) error {
+func (system *System) SaveClientToken(label string, xData string) (string, error) {
```
dhgreene, line 93: Change "toke" to "token" above

dhgreene (approved):
🍪

PR #84, opened 2021-06-21
DPC-1463 Enabling inline private keys as environment vars

orenfromberg:
### Related to [DPC-1463](https://jira.cms.gov/browse/DPC-1463)

This branch enables inline private keys as environment variables for the public and admin signing keys.

This feature is needed for deploying SSAS as a container since the private keys will not be available in the container file system and will need to be supplied via environment variables.

### Proposed Changes

Adding two new environment variables `SSAS_PUBLIC_SIGNING_KEY` and `SSAS_ADMIN_SIGNING_KEY` that will preempt `SSAS_PUBLIC_SIGNING_KEY_PATH` and `SSAS_ADMIN_SIGNING_KEY_PATH`, respectively.

### Change Details

Additionally, I've updated the unit test to generate a private key at run time instead of reading one from the file system.

### Security Implications

### Acceptance Validation

- [x] public and admin signing keys can be specified with plaintext in environment variables.

### Feedback Requested

All feedback is welcome. There is some duplicated code so I'd like a suggestion of how to make it more DRY.


msnook:
I added @CosmicSyntax as a reviewer as well.

ssas/service/admin/router.go
```diff
@@ -16,17 +18,40 @@ import (
...
 }
 
 // Server creates an SSAS admin server
 func Server() *service.Server {
 	unsafeMode := os.Getenv("HTTP_ONLY") == "true"
-	server = service.NewServer("admin", ":3004", constants.Version, infoMap, routes(), unsafeMode, adminSigningKeyPath, 20*time.Minute, "")
+
+	var signingKey *rsa.PrivateKey
+
+	if adminSigningKey != "" {
```
msnook, line 36: Should we add comments in the code, prior to the conditional, to explain whats going on?  
(The same comment applies in `public/router.go`)

msnook, line 36: To reduce the possibility of confusion, should we enforce that only ONE of either the key or the path can be set?  In the current implementation, both values could be set (to potentially different values), but the code will only pick up the key value and ignore the path value.  Should it be considered a server init/startup error if both are set? 
(The same comment applies in `public/router.go`)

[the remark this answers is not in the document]
orenfromberg: I added test coverage for both `ChooseSigningKey()` and `NewServer()`.

orenfromberg: I was planning on using  the inline environment variable as an override but this is reasonable. I fixed it so that it will fail if both or none are set, and will only accept one or the other.

msnook (approved):
👍

CosmicSyntax (approved):
🎉

dhgreene (approved):
👍

PR #88, opened 2021-06-25
DPC-1388 return system information

MrBilnon:
### Fixes [DPC-1388](https://jira.cms.gov/browse/DPC-1388)

Create a get endpoint for v2 system and return data about the system

### Change Details

* create swagger output structs that is a subset of the data from the databases
* use those swagger output structs in the response of api

### Security Implications

### Acceptance Validation

Screen Shot 2021-06-25 at 8 40 48 PM

### DIFF
--- a/ssas/service/admin/api.go
+++ b/ssas/service/admin/api.go
@@ -192,13 +192,6 @@ func getSystem(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if s.DeletedAt.Valid {
-		systemEvent.Help = fmt.Sprintf("system %s has been deleted", id)
-		ssas.OperationFailed(systemEvent)
-		jsonError(w, http.StatusNotFound, fmt.Sprintf("could not find system %s", id))
-		return
-	}
-
 	ips, _ := s.GetIPsData()
 	cts, _ := s.GetClientTokens(trackingID)
 	eks, _ := s.GetEncryptionKey(trackingID)
--- a/ssas/service/admin/api.go
+++ b/ssas/service/admin/api.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/CMSgov/bcda-ssas-app/ssas"
 	"github.com/CMSgov/bcda-ssas-app/ssas/service"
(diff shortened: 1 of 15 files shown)


ssas/systems.go
```diff
@@ -82,9 +82,9 @@ type ClientToken struct {
 	gorm.Model
 	Label     string    `json:"label"`
 	Uuid      string    `json:"uuid"`
-	System    System    `gorm:"foreignkey:SystemID;association_foreignkey:ID"`
-	SystemID  uint      `json:"system_id"`
-	ExpiresAt time.Time `json:"expires_at"`
+	System    System    `gorm:"foreignkey:SystemID;association_foreignkey:ID" json:"-"`
+	SystemID  uint      `json:"-"`
+	ExpiresAt time.Time `json:"-"`
```
jonfulk, line 87: we don't want to expose when it expires? Is that for security purposes? It seems like it could be helpful information.

jonfulk (approved):
LGTM

ssas/service/admin/api.go
```diff
@@ -178,6 +179,52 @@ func updateGroup(w http.ResponseWriter, r *http.Request) {
...
+	systemEvent := ssas.Event{Op: "UpdateSystem", TrackingID: trackingID}
+
+	s, err := ssas.GetSystemByID(id)
+	if err != nil {
+		systemEvent.Help = fmt.Sprintf("could not find system %s", id)
+		ssas.OperationFailed(systemEvent)
+		jsonError(w, http.StatusNotFound, fmt.Sprintf("could not find system %s", id))
+		return
+	}
+
```
dhgreene, line 194: We probably ought to add a verification that this system is active

ssas/systems.go
dhgreene: @MrBilnon care to add client token expiration date?

ssas/service/admin/api.go
MrBilnon: dobe

ssas/systems.go
MrBilnon: dobe

ssas/service/admin/api.go
dhgreene: Can we also add a unit test case?

MrMorie (dismissed):
LGTM

dhgreene (approved):
💯

PR #91, opened 2021-07-05
DPC-1560 post and delete token

MrBilnon:
### Fixes [DPC-1560](https://jira.cms.gov/browse/DPC-1560)

In SSAS v2, we need to be able to create and delete (multiple) client tokens in a system.

### Change Details

Added two methods, one to delete and one to create, the create method uses an existing method from when we registered a v2 system and the delete method uses gorm to retrieve the root key by token uuid and system id and then deletes both the root key and client token associated with that root key.  All deletes are soft deletes.

### Security Implications

### Feedback Requested

* Should there be a label length limit?

### DIFF
--- a/ssas/service/admin/api.go
+++ b/ssas/service/admin/api.go
@@ -700,7 +700,7 @@ func createToken(w http.ResponseWriter, r *http.Request) {
 
 	if body["label"] == "" {
 		ssas.OperationFailed(tokenEvent)
-		jsonError(w, http.StatusInternalServerError, "Internal Error")
+		jsonError(w, http.StatusBadRequest, "Missing label")
 		return
 	}
 
@@ -708,7 +708,6 @@ func createToken(w http.ResponseWriter, r *http.Request) {
 	ct, m, err := system.SaveClientToken(body["label"], group.XData, expiration)
 	if err != nil {
 		tokenEvent.Help = fmt.Sprintf("could not save client token for clientID %s, groupID %s: %s", system.ClientID, system.GroupID, err.Error())
-		ssas.OperationFailed(tokenEvent)
 		ssas.OperationFailed(tokenEvent)
 		jsonError(w, http.StatusInternalServerError, "Internal Error")
 	}
@@ -721,7 +720,6 @@ func createToken(w http.ResponseWriter, r *http.Request) {
 	b, err = json.Marshal(response)
 	if err != nil {
 		tokenEvent.Help = fmt.Sprintf("could not marshal response for clientID %s, groupID %s: %s", system.ClientID, system.GroupID, err.Error())
-		ssas.OperationFailed(tokenEvent)
 		ssas.OperationFailed(tokenEvent)
 		jsonError(w, http.StatusInternalServerError, "Internal Error")
 	}
--- a/ssas/service/admin/api.go
+++ b/ssas/service/admin/api.go
@@ -663,6 +663,86 @@ func deleteSystemIP(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusNoContent)
 }
 
+func createToken(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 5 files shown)


ssas/service/admin/api.go
```diff
@@ -663,6 +663,99 @@ func deleteSystemIP(w http.ResponseWriter, r *http.Request) {
...
+
+	if err := json.Unmarshal(b, &body); err != nil {
+		ssas.OperationFailed(tokenEvent)
+		jsonError(w, http.StatusInternalServerError, "Internal Error")
+		return
+	}
+
+	if body["label"] == "" {
+		ssas.OperationFailed(tokenEvent)
+		jsonError(w, http.StatusInternalServerError, "Internal Error")
```
dhgreene, line 703: It doesn't feel like an internal error if required input is missing

ssas/service/admin/api.go
```diff
@@ -663,6 +663,99 @@ func deleteSystemIP(w http.ResponseWriter, r *http.Request) {
...
+	response := ssas.ClientTokenResponse{
+		ClientTokenOutput: ssas.OutputCT(*ct)[0],
+		Token:             m,
+	}
+
+	b, err = json.Marshal(response)
+	if err != nil {
+		tokenEvent.Help = fmt.Sprintf("could not marshal response for clientID %s, groupID %s: %s", system.ClientID, system.GroupID, err.Error())
+		ssas.OperationFailed(tokenEvent)
+		ssas.OperationFailed(tokenEvent)
```
dhgreene, line 725:
```suggestion
```

ssas/service/admin/api.go
```diff
@@ -663,6 +663,99 @@ func deleteSystemIP(w http.ResponseWriter, r *http.Request) {
...
+		jsonError(w, http.StatusInternalServerError, "Internal Error")
+		return
+	}
+
+	expiration := time.Now().Add(ssas.MacaroonExpiration)
+	ct, m, err := system.SaveClientToken(body["label"], group.XData, expiration)
+	if err != nil {
+		tokenEvent.Help = fmt.Sprintf("could not save client token for clientID %s, groupID %s: %s", system.ClientID, system.GroupID, err.Error())
+		ssas.OperationFailed(tokenEvent)
+		ssas.OperationFailed(tokenEvent)
```
dhgreene, line 712:
```suggestion
```

ssas/service/admin/api.go
```diff
@@ -663,6 +663,99 @@ func deleteSystemIP(w http.ResponseWriter, r *http.Request) {
...
+
+	system, err := ssas.GetSystemByID(systemID)
+	if err != nil {
+		ssas.OperationFailed(tokenEvent)
+		jsonError(w, http.StatusNotFound, "Invalid system ID")
+		return
+	}
+
+	ssas.OperationCalled(tokenEvent)
+	err = system.DeleteClientToken(tokenID)
```
dhgreene, line 749: Should we verify that the token belongs to the system?

ssas/systems.go
```diff
@@ -135,6 +135,28 @@ func (system *System) GetClientTokens(trackingID string) ([]ClientToken, error)
...
+func (system *System) DeleteClientToken(tokenID string) error {
+	db := GetGORMDbConnection()
+	defer Close(db)
+
+	tx := db.Begin()
+	defer func() {
+		if r := recover(); r != nil {
+			tx.Rollback()
+		}
+	}()
```
dhgreene, line 147: When there's just one SQL statement, do we gain anything by using a transaction?

MrBilnon: @dhgreene it's two sql statements though

dhgreene: But it doesn't have to be 😈

dhgreene: On third reading, I see that you're wanting to delete both the root key and the client token. 👍

dhgreene (approved):
🦖

PR #92, opened 2021-07-07
DPC-1561 post and delete key

MrBilnon:
### Fixes [DPC-1561](https://jira.cms.gov/browse/DPC-1561)

In SSAS v2, we need to be able to register and delete (multiple) public keys in a system.

### Proposed Changes

* created two endpoints
* one endpoint to create a key by using existing savePublicKey code
* one endpoint to delete a key by creating a new function to delete a key using gorm
* Ability to save multiple keys per system
* Ability to verify a signature like current DPC v1

### Security Implications

### DIFF
--- a/ssas/service/admin/api.go
+++ b/ssas/service/admin/api.go
@@ -663,6 +663,60 @@ func deleteSystemIP(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusNoContent)
 }
 
+func createKey(w http.ResponseWriter, r *http.Request) {
+	systemID := chi.URLParam(r, "systemID")
+	trackingID := ssas.RandomHexID()
+	keyEvent := ssas.Event{Op: "CreateKey", TrackingID: trackingID, Help: "calling from admin.createKey()"}
+
+	system, err := ssas.GetSystemByID(systemID)
+	if err != nil {
+		ssas.OperationFailed(keyEvent)
+		jsonError(w, http.StatusNotFound, "Invalid system ID")
+		return
+	}
+
+	var pk ssas.PublicKeyInput
+	if err := json.NewDecoder(r.Body).Decode(&pk); err != nil {
+		ssas.OperationFailed(keyEvent)
+		jsonError(w, http.StatusBadRequest, "Failed to read body")
+		return
+	}
+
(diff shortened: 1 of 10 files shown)


ssas/service/admin/api.go
```diff
@@ -663,6 +663,60 @@ func deleteSystemIP(w http.ResponseWriter, r *http.Request) {
...
+		ssas.OperationFailed(keyEvent)
+		jsonError(w, http.StatusBadRequest, "Failed to read body")
+		return
+	}
+
+	if err := system.SavePublicKey(strings.NewReader(pk.PublicKey)); err != nil {
+		ssas.OperationFailed(keyEvent)
+		jsonError(w, http.StatusInternalServerError, "Failed to save public key")
+		return
+	}
```
dhgreene, line 689: How difficult would it be to 1) optionally accept the signed text that DPC currently does 2) if present, validate the signature, rejecting the key if the signature's not valid?

MrBilnon: @dhgreene probably not too difficult, looking at the Java implementation, it seems pretty straight forward. Will just need to research and find equivalent things to Java

MrMorie (dismissed):
LGTM

ssas/service/admin/api.go
```diff
@@ -754,6 +754,61 @@ func deleteToken(w http.ResponseWriter, r *http.Request) {
...
+		ssas.OperationFailed(keyEvent)
+		jsonError(w, http.StatusInternalServerError, "Failed to add additional public key")
+		return
+	}
+
+	key, _ := system.GetEncryptionKey(trackingID)
+
+	w.Header().Set("Content-Type", "application/json")
+	keyStr := strings.Replace(key.Body, "\n", "\\n", -1)
+	fmt.Fprintf(w, `{ "client_id": "%s", "public_key": "%s" }`, system.ClientID, keyStr)
```
dhgreene, line 786: We won't need it (at least right away, so I won't insist), but it's a good practice to output the ID of the object that was created. In `deleteKey()` the user needs it.

dhgreene (commented):
Are there test cases for `onlyOne` with true/false values?

dhgreene (approved):
🔑

PR #96, opened 2021-08-17
DPC-1684 Used KID to lookup encryption key. (POST /v2/token)

MrMorie:
### Fixes [DPC-1684](https://jira.cms.gov/browse/DPC-1684)

During a SSAS debugging of POST /v2/token session with Jonathan Fulk and William Huang it was discovered that SSAS was always using the the first public key (uploaded during system creation) to verify the client assertion.

This is an issue when more than one public key is uploaded, the client assertion could be signed with a second private key (linked to the second public key), but the first public key would be used.

This is due to SSAS looking up the encryption key by system id instead of using the 'kid' specified in the signed JWT

It was also discovered that the UUID of the public key created during system creation was not being returned. Not a big issue since the web portal (our only user of POST /v2/System) is not using this value, instead it is calling GET /v2/system to retrieve all public keys and ids.

### Proposed Changes
SSAS POST /v2/Token should use the jwt's 'kid' found in the tokens header.

### Change Details

- Added new method to lookup encryption key by UUID
- Returned 'public_key_id' when a new system is created.
- Used the encryption key specified in JTW kid

### Security Implications

No new software dependencies.

- [x] security controls or supporting software altered
Yes, to fix a bug. Previously only the first encryption would be used, now the client may specify what public to to use for assertion verification (public key must belong to client/system)

No new data is stored or transmitted.

No, a security check list was not completed for this specific change. This is a bugfix for an existing authentication mechanism specified in the SSAS V2 Security checklist: https://confluence.cms.gov/pages/viewpage.action?spaceKey=DAPC&title=2021-04-29%3A+SSAS+v2+Security+Checklist

- [x] requires more information or team discussion to evaluate security implications
Is a security check list required fo this change?

No PHI/PII related changes.

### Acceptance Validation

Yes, via unit test.

Screen Shot 2021-09-01 at 6 17 04 PM

### Feedback Requested
- Do we need a security check list for this change?
- Does web app need to make any changes?

### DIFF
--- a/ssas/service/admin/api_test.go
+++ b/ssas/service/admin/api_test.go
@@ -496,13 +496,13 @@ func (s *APITestSuite) TestGetPublicKeyRotation() {
 		s.FailNow(err.Error())
 	}
 
-	key1, _, _ := ssas.GeneratePublicKey(2048)
+	key1, _, _, _ := ssas.GeneratePublicKey(2048)
 	rk1, err := system.SavePublicKey(strings.NewReader(key1), "")
 	if err != nil {
 		s.FailNow(err.Error())
 	}
 
-	key2, _, _ := ssas.GeneratePublicKey(2048)
+	key2, _, _, _ := ssas.GeneratePublicKey(2048)
 	rk2, err := system.SavePublicKey(strings.NewReader(key2), "")
 	if err != nil {
 		s.FailNow(err.Error())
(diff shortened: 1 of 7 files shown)


dhgreene (approved):
Thanks for finding and fixing these by using the right tool in the standard!

MrBilnon (approved):

MrBilnon (approved):

PR #105, opened 2022-01-04
DPC-2036: Update Logfile permissions

dhgreene:
### Fixes [DPC-2036](https://jira.cms.gov/browse/DPC-2036)
DPC is experiencing issues with Splunk intermittently refusing to continue ingesting SSAS logs due to lack of permissions. This PR proposes to set SSAS logfile permissions to the same level used by other services.

### Proposed Changes
- Set log permissions to allow group writes and universal reads
- Update README examples

### Security Implications
- [x] security checklist is completed for this change
[Security checklist](https://confluence.cms.gov/display/DAPC/2022-01-04%3A+SSAS+log+permissions+Security+Checklist)

Note that this PR:
1. Updates the file permissions to match those in use by other services using SSAS
2. Is expected to fix an intermittent (but frequent) issue where Splunk stops ingesting logs

#### ACO API tests pass
Screen Shot 2022-01-04 at 1 10 26 PM

##### Top entry is a different service's logfile; Middle entry is new SSAS logfile; bottom entry is a previous SSAS logfile
Screen Shot 2022-01-04 at 2 02 53 PM

### Feedback Requested
Any concerns?

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,13 +140,13 @@ environmental variables. It is also possible to run individual tests, but that m
 docker run --rm --network bcda-ssas-app_default -it postgres pg_dump -s -h bcda-ssas-app_db_1 -U postgres bcda > schema.sql
 ```
 ```
-docker-compose run --rm ssas sh -c 'main --reset-secret --client-id=[client_id]'
+docker-compose run --rm ssas sh -c 'ssas --reset-secret --client-id=[client_id]'
 ```
 ```
-docker-compose run --rm ssas sh -c 'main --list-ips'
+docker-compose run --rm ssas sh -c 'ssas --list-ips'
 ```
 ```
-docker-compose run --rm ssas sh -c 'main --new-admin-system --system-name=[entity name]'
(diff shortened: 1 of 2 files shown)


orenfromberg (approved):
If I'm reading this correctly, it will give write permission to users in the logreader group. Does that include the user that is running the splunk ingestor? What could it be doing with that permission? I'm not sure if we want to do this or why it would work, but if other logs seem to be ingested correctly then it's worth trying. /shrug

dhgreene:
> If I'm reading this correctly, it will give write permission to users in the logreader group. Does that include the user that is running the splunk ingestor? What could it be doing with that permission? I'm not sure if we want to do this or why it would work, but if other logs seem to be ingested correctly then it's worth trying. /shrug

Specifically we're facing the following error in `splunkd.log`:
```
 -0500 ERROR TailReader - Ignoring path="/var/log/dpc/dpc-ssas-ip-10-247-2-176.ec2.internal.log" due to:   Bug during applyPendingMetadata, header processor does not own the indexed extractions confs.
```

Honestly, it's hard to prove that the changes in this PR address that problem until we don't see that message any more for a long time.

PR #120, opened 2022-11-10
elle/BCDA-6289: ValidateClientIdAndSecret function added, added enhanced logging

4ell0:
### Change Details
1. err != nil block when error happens during GetSystemByClientId to include a log on what the actual error is so that it can be picked up in Splunk
2. err != nil block  during GetSecret to include a log on what the actual error is so that it can be picked up in Splunk,. Additionally, also log result of assertion in case the block was reached due to mismatch of hash comparison.
3. update where JSONError logged if MintAccessToken reached an error to have statusDescription updated from empty space "" to "failure minting token"
4Refactor cleanup: Can extract lines 479-500 in a separate method 

### Security Implications


ssas/service/public/api_test.go
```diff
@@ -323,6 +326,141 @@ func (s *APITestSuite) TestTokenSuccess() {
...
+	handler.ServeHTTP(s.rr, req)
+
+	assert.Equal(s.T(), http.StatusUnauthorized, s.rr.Code)
+	assert.Contains(s.T(), s.rr.Body.String(), "invalid client id")
+
+	err = ssas.CleanDatabase(group)
+	assert.Nil(s.T(), err)
+}
+
+// Test is error is  not retrieved
```
krobertson3314, line 433: can we remove this commented out block of code? This was from a while back when initially making the code changes and I'm not sure what this test is trying to accomplish here.

PR #131, opened 2023-05-11
Increase Get Access Token request timeout

laurenkrugen-navapbc:
## 🎫 Ticket

<>

## 🛠 Changes

Upping the timeout function to wait for 3 s instead of 300 ms during the `public Get Access Token` smoke test. 

## ℹ️ Context for reviewers

Smoke test pipeline will pass/fail intermittently, typically on the test to retrieve a token. This delays deployments and also makes complicates the addition of new smoke tests.

## ✅ Acceptance Validation

To be tested in jenkins, no negative impacts with the exception of a 3 second longer run time.

## 🔒 Security Implications

Submitters should complete the following questionnaire:

If any of the statements above were checked, then please add Jason Ashbaugh (GitHub username: StewGoin) as a reviewer, and note that this PR should not be merged unless/until he also approves it.

### DIFF
--- a/test/postman_test/SSAS_Smoke_Test.postman_collection.json
+++ b/test/postman_test/SSAS_Smoke_Test.postman_collection.json
@@ -2532,7 +2532,7 @@
 									"postman.setEnvironmentVariable(\"jwt_token\", sJWT);",
 									"",
 									"// slow down this request to ensure token is not \"used before issued\"",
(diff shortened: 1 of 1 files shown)


devopsmatt (approved):
LGTM

devopsmatt (approved):
LGTM

PR #133, opened 2023-05-26
BCDA-7085: Use a single connection pool for GORM with limits

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7085

## 🛠 Changes

- Switch to global `Connection` object which represents a GORM connection pool that's instantiated automatically by the `init` method in connection.go
- Upgrade gorm and gorm postgres driver to support usage of custom pgx DB connection
- Add limitations on number of connections in the pool, and the max idle timeout

## ℹ️ Context for reviewers

The BCDA application caps the number of allowed DB connections to 40, and uses a single connection pool. SSAS creates a new pool for every query and doesn't cap the number of connections.

Although this doesn't appear to have a major performance impact, it does seem like it could be the root cause of queries slowing down by 100-200ms when there are many concurrent API requests to SSAS.By limiting the number of queries running concurrently, we can reduce database load and hopefully keep individual query times down.

In addition, we have observed during performance testing that the SSAS app can hit the postgres "too many database connections" error, which will either cause the application to crash or return a 503 Service Unavailable error.

By capping the number of database connections, the application will know to wait for a new database connection to become available, so API requests can receive a successful response (with some time delay.)

## ✅ Acceptance Validation

- I verified that unit and local postman tests ran successfully.
- I verified that I could hit the API /token endpoint immediately after starting the application, and after waiting overnight (to ensure that new DB connections are made appropriately after the 5-min idle timeout)

I also verified that when the database goes down (`docker-compose stop db`), I can see error logs and an HTTP error response from SSAS:

```
bcda-ssas-app-ssas-1               | 2023/05/26 14:02:22 /go/src/github.com/CMSgov/bcda-ssas-app/ssas/systems.go:827 failed to connect to `host=db user=postgres database=bcda`: hostname resolving error (lookup db on 127.0.0.11:53: no such host)
bcda-ssas-app-ssas-1               | [34.669ms] [rows:0] SELECT * FROM "systems" WHERE client_id = '0c527d2e-2e8a-4808-b11d-0fa06baf8254' AND "systems"."deleted_at" IS NULL ORDER BY "systems"."id" LIMIT 1
bcda-ssas-app-ssas-1               | {"level":"error","msg":"The client id failed to connect to `host=db user=postgres database=bcda`: hostname resolving error (lookup db on 127.0.0.11:53: no such host) is invalid","time":"2023-05-26T14:02:22.937185295Z"}
bcda-ssas-app-ssas-1               | {"level":"info","msg":"Unauthorized; invalid client id","time":"2023-05-26T14:02:22.937826628Z"}
```

And when I bring the database back up (`docker-compose up db`), I can call SSAS successfully:

```
$ curl -X POST -u CLIENT_ID:CLIENT_SECRET localhost:3003/token
{"access_token":"..."}
```

## What this doesn't fix

After performing a locust test with these changes in my local environment, I've concluded that this change, by itself, is not enough to resolve our SSAS uptime issues. With the new connection pool, we successfully keep the service _running_ instead of crashing. 

**This is good!**

However, when requests start to time out, **they do not release database connections back into the pool**. When printing out the Connection stats, it will eventually indicate that all connections are in use even though in the DB they are idle. At this point, all requests to the SSAS API will fail even if there are no other concurrent requests being made.

To resolve this, we will need a followup ticket to ensure that database connections eventually timeout, by providing a cancellable context to each database query call (similar to BCDA.) In practice, this probably means passing `r.Context()` into our db calls so that database queries are canceled when the request is canceled.

This will ensure that even when DB queries are taking a long time, they will listen to the timeout events and release themselves back into the pool for future requests as we intended.

## Additional Notes

- You'll notice in the logs above that the response to the user when we can't connect to the DB is "401: unauthorized - invalid client id". They should receive a 503 instead. I will create a separate ticket for this.

## 🔒 Security Implications


db/migrations/migrations_test.go
```diff
@@ -1,4 +1,4 @@
-// +build migrations
+//go:build migrations
```
kyeah, line 1: This is a change in directive style introduced in go 1.18:

https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives

This was an automated change from our pre-commit hook.

ssas/blacklist.go
```diff
@@ -32,9 +33,7 @@ func CreateBlacklistEntry(key string, entryDate time.Time, cacheExpiration time.
 		CacheExpiration: cacheExpiration.UnixNano(),
 	}
 
-	db := GetGORMDbConnection()
-	defer Close(db)
-	err = db.Save(&be).Error
+	err = Connection.Save(&be).Error
```
kyeah, line 36: It's not super clear where `Connection` comes from since it lives in the top-level `ssas` package. I'll create a ticket to move the packages around so we can say `db.Connection`.

ssas/connection.go
```diff
@@ -1,35 +1,53 @@
...
+
+	// db.SetMaxOpenConns(cfg.MaxOpenConns)
+	// db.SetMaxIdleConns(cfg.MaxIdleConns)
+	// db.SetConnMaxLifetime(time.Duration(cfg.ConnMaxLifetimeMin) * time.Minute)
+
+	sqlDB.SetMaxOpenConns(40)
+	sqlDB.SetMaxIdleConns(40)
+	sqlDB.SetConnMaxLifetime(time.Duration(5) * time.Minute)
+
+	db, err := gorm.Open(postgres.New(postgres.Config{
```
kyeah, line 40: `postgres.New` and `postgres.Config` are only supported in newer versions of the gorm postgres driver, so I had to upgrade. This also seems to force an upgrade of gorm.

go.mod
```diff
@@ -11,32 +11,31 @@ require (
...
 	github.com/pborman/uuid v1.2.1
 	github.com/sirupsen/logrus v1.9.0
-	github.com/stretchr/testify v1.8.0
-	golang.org/x/crypto v0.0.0-20220826181053-bd7e27e6170d
+	github.com/stretchr/testify v1.8.1
+	golang.org/x/crypto v0.8.0
 	gopkg.in/macaroon.v2 v2.1.0
-	gorm.io/driver/postgres v1.3.9
-	gorm.io/gorm v1.23.8
+	gorm.io/driver/postgres v1.5.2
```
kyeah, line 17: Changelog: https://github.com/go-gorm/postgres/tags

No breaking changes or other worrying things in there from what I can tell. Lots of very specific bug fixes.

go.mod
```diff
@@ -11,32 +11,31 @@ require (
...
+	gorm.io/gorm v1.25.0
```
kyeah, line 18: This change is a bit of a jump because v1.23.8 was released Sept 2022. There were two function definition changes that I needed to update our code to (the Golang compiler caught them.)

https://github.com/go-gorm/gorm/tags

A little hard to guarantee full correctness but I'm putting some of my faith in the compiler + the test coverage to ensure everything is working correctly 😬

go.mod
```diff
@@ -11,32 +11,31 @@ require (
...
+	golang.org/x/net v0.9.0 // indirect
+	golang.org/x/sys v0.7.0 // indirect
+	golang.org/x/text v0.9.0 // indirect
 	google.golang.org/genproto v0.0.0-20220822174746-9e6da59bd2fc // indirect
 	google.golang.org/grpc v1.49.0 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
-
-replace (
```
kyeah, line 39: I'm not sure why these replacements were here.

ssas/connection.go
```diff
@@ -1,35 +1,53 @@
...
-	databaseURL := os.Getenv("DATABASE_URL")
-	db, err := gorm.Open(postgres.Open(databaseURL), &gorm.Config{})
 	if err != nil {
-		LogFatal(err)
+		Logger.Fatalf("Failed to create db %s", err.Error())
 	}
-	return db
 }
 
-func Close(db *gorm.DB) {
```
kyeah, line 25: We don't need this method anymore because we never want to forcefully close the gorm DB pool. `Connection` will self-manage the creation and removal of database connections based on the configuration (MaxOpenConns, MaxIdleConns, ConnMaxLifetime)

ssas/service/main/main.go
```diff
@@ -190,8 +190,7 @@ func newForwardingRouter() http.Handler {
 }
 
 func addFixtureData() {
-	db := ssas.GetGORMDbConnection()
-	defer ssas.Close(db)
+	db := ssas.Connection
```
kyeah, line 193: We could probably inline `ssas.Connection` in all of these tests instead of assigning a new variable, but I wanted to keep the changes in this PR as minimal as possible for clarity. I'll follow up afterwards with some test cleanup.

ssas/systems_test.go
```diff
@@ -434,7 +430,7 @@ func (s *SystemsTestSuite) TestSystemClientGroupDuplicate() {
 
 	system = System{GID: group2.ID, ClientID: "498765uzyxwv", ClientName: "Duplicate Client"}
 	err = s.db.Create(&system).Error
-	assert.EqualError(err, "pq: duplicate key value violates unique constraint \"idx_client\"")
+	assert.EqualError(err, "ERROR: duplicate key value violates unique constraint \"idx_client\" (SQLSTATE 23505)")
```
kyeah, line 433: This error message changed slightly with the gorm/postgres driver upgrade.

ssas/service/api_common_test.go
```diff
@@ -2,14 +2,15 @@ package service
...
-	"github.com/CMSgov/bcda-ssas-app/ssas"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/suite"
-	"gorm.io/gorm"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+
+	"github.com/CMSgov/bcda-ssas-app/ssas"
```
laurenkrugen-navapbc, line 10: like the import ordering! 🎉

ssas/service/main/main.go
laurenkrugen-navapbc: Appreciate keeping the PR small since this update touched a lot of files. Will keep that in mind for the next PR! 👍

ssas/groups.go
```diff
@@ -143,8 +136,6 @@ func DeleteGroup(id string) error {
 	event := Event{Op: "DeleteGroup", TrackingID: id}
 	OperationStarted(event)
 
-	db := GetGORMDbConnection()
```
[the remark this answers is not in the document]
kyeah: Nope. It's been a topic of discussion from past teams about moving to a singleton connection pool, but they didn't have concrete reason to do so previously. Our latest performance tests indicate that doing this will help us avoid the "too many connections" error from postgres (along with matching the pattern established in the main BCDA application.)

ssas/service/api_common_test.go
kyeah: my IDE does this automatically 😂 can't help it 🤷

go.mod
laurenkrugen-navapbc: got it! I also did some looking and didn't see anything immediately concerning in terms of compatibility/changes.

laurenkrugen-navapbc (approved):

PR #135, opened 2023-05-31
BCDA-7113: Ensure database connections are returned when request times out

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7113

## 🛠 Changes

- Run database queries with cancellable request context when possible

## ℹ️ Context for reviewers

In #133 we are switching to a singleton DB connection pool. However, during performance testing, we noticed that connections were never returned to the pool if requests timed out.

To avoid this, we need to pass a cancelable context into each database query call. This PR does the following:

- For every database call (`Connection.X`), swap it out with `Connection.WithContext(ctx)` if it is ever called during an API request. 
- For database calls that are _only_ used during unit testing or CLI commands, don't do anything.
- For every instance where we need a context, update the function parameters to accept a context object. This will usually be a `r.Context()` in requests, and `context.Background()` (empty context) in tests.

Note that in rare scenarios, we are running within an "API" context but don't have the request context available. In these 2 cases, I created a new context with a hardcoded timeout of 5 seconds.

## ✅ Acceptance Validation

[See the concurrency test results here](https://jira.cms.gov/browse/BCDA-7113?focusedId=3199142&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3199142).

## 🔒 Security Implications


ssas/groups.go
```diff
@@ -178,35 +179,28 @@ func GetAuthorizedGroupsForOktaID(oktaID string) ([]string, error) {
...
+		if r := recover(); r != nil {
+			tx.Rollback()
+		}
+	}()
 
-	err = Connection.Where("id IN (?)", systemIds).Delete(&system).Error
-	if err != nil {
-		return fmt.Errorf("unable to delete systems: %s", err.Error())
-	}
+	tx.Table("systems").Where("group_id = ?", group.GroupID).Pluck("id", &systemIds)
```
laurenkrugen-navapbc, line 197: the use of transaction simplifies the error handling nicely in this section. 🎉

ssas/groups.go
```diff
@@ -257,21 +258,21 @@ type Resource struct {
...
 
-func GetGroupByGroupID(groupID string) (Group, error) {
+func GetGroupByGroupID(ctx context.Context, groupID string) (Group, error) {
 	var (
 		group Group
 		err   error
 	)
 
-	if err = Connection.First(&group, "group_id = ?", groupID).Error; err != nil && errors.Is(err, gorm.ErrRecordNotFound) {
+	if err = Connection.WithContext(ctx).First(&group, "group_id = ?", groupID).Error; err != nil && errors.Is(err, gorm.ErrRecordNotFound) {
```
laurenkrugen-navapbc, line 267: In some places that uses `.first()`, we check for any error that is returned and here we specifically check for the `ErrRecordNotFound` error. Should this be a more generic, catch-all error handling here or do you think this specific error is more appropriate? 🤔

ssas/service/tokenblacklist.go
```diff
@@ -32,12 +32,12 @@ func init() {
 
 // This function should only be called by main
 func StartBlacklist() {
-	TokenBlacklist = NewBlacklist(defaultCacheTimeout, cacheCleanupInterval)
+	TokenBlacklist = NewBlacklist(context.Background(), defaultCacheTimeout, cacheCleanupInterval)
 }
 
 //	NewBlacklist allows for easy Blacklist{} creation and manipulation during testing, and, outside a test suite,
 //	should not be called
```
laurenkrugen-navapbc, line 39: this is outside the scope of this PR but there seems a trend of test fixtures that aren't within test files. Would it we worthwhile to discuss how we can clean these up?

laurenkrugen-navapbc (approved):

ssas/service/tokenblacklist.go
kyeah: yeah I noticed that as well. We should build up a list and decide how we can move them into a test utilities file or something.

ssas/groups.go
kyeah: We actually return the error below even if it's not an ErrRecordNotFound error, so it is handling all errors. I notice in a couple places we don't actually log or use the error object at all though 😅 Like this API method doesn't use the err:

https://github.com/CMSgov/bcda-ssas-app/blob/9d851236c60174d5e0679b1d3bf1d6792c9c081b/ssas/service/admin/api.go#L680-L682

laurenkrugen-navapbc: doh, I must have looked right past that. I noticed your second point too! lots of place where that actual err object is not being used..

PR #139, opened 2023-06-21
BCDA-7022: Add standardized log attributes to SSAS application

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7022

## 🛠 Changes

Addition of standarized log attributes (environment, application, and version) to all log events that are generated.

## ℹ️ Context for reviewers

Part of ongoing effort to improve logging in the SSAS/API projects.

## ✅ Acceptance Validation

Ran locally, verified stdout.

## 🔒 Security Implications


ssas/logger.go
```diff
@@ -22,48 +23,58 @@ type Event struct {
...
+}
+
+type APILoggerEntry struct {
+	Logger logrus.FieldLogger
 }
 
-func mergeNonEmpty(data Event) *logrus.Entry {
-	var entry = logrus.NewEntry(Logger)
+func mergeNonEmpty(data Event) logrus.FieldLogger {
+	entry := &APILoggerEntry{Logger: Logger}
```
kyeah, line 56: Does this still need to create a new entry or is it okay to re-use the existing FieldLogger? This seems okay but I don't really know the difference 😅

ssas/service/logging.go
```diff
@@ -13,18 +13,19 @@ import (
...
 
 type APILogger struct {
-	Logger *logrus.Logger
+	Logger logrus.FieldLogger
 }
 
 func (l *APILogger) NewLogEntry(r *http.Request) middleware.LogEntry {
-	entry := &APILoggerEntry{Logger: logrus.NewEntry(l.Logger)}
+	// change the parameter to take an entry for fieldLogger, change line 23 to entry, pass in an entry
+	entry := &APILoggerEntry{Logger: l.Logger}
```
kyeah, line 28: similar question to the above — is there a difference between using the existing logger vs. creating a new entry?

ssas/service/public/api_test.go
```diff
@@ -460,17 +461,25 @@ func (s *APITestSuite) TestTokenEmptyClientIdProduces401() {
 	assert.Nil(s.T(), err)
 }
 
+func GetLogger(logger logrus.FieldLogger) *logrus.Logger {
```
kyeah, line 464: can this be imported from the testutils?

ssas/logger.go
```diff
@@ -22,48 +23,58 @@ type Event struct {
...
+			logInstance.Info("Failed to open SSAS log file; using default stderr")
 		}
 	} else {
-		Logger.Info("No SSAS log location provided; using default stderr")
+		logInstance.Info("No SSAS log location provided; using default stderr")
 	}
+
+	Logger = logInstance.WithFields(logrus.Fields{
+		"application": constants.Application,
+		"environment": os.Getenv("DEPLOYMENT TARGET"),
```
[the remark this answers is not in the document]
kyeah: We should also add this variable to docker-compose.yml if it doesn't already exist there for local development 😄

ssas/service/public/api_test.go
laurenkrugen-navapbc: another good catch - I was testing out this function in a single file and forgot to remove it here. thank you!

ssas/service/logging.go
laurenkrugen-navapbc: this was an odd hurdle; creating a New Entry will remove the env, app, and version fields. when I ran it locally, it didn't merge all of the fields, but I can do some additional testing to make sure there's no off cases where that might happen. Another alternative would be to create a logger instance for the middleware and one for the operations.

laurenkrugen-navapbc: when we create log fields and use `withField` or `withFields`, it creates a new Entry object, so this entry for the middleware and the entry for the operations should stay separate. Using either of those functions doesn't attribute the fields to the logger, but the returned Entry can be used to persist fields during log calls -- seems a little counterintuitive. The use of FieldLogger generalizes Entry and Logger which is how we can pass around the instance.

kyeah: thanks for the detailed explanation! The fact that FieldLogger is a generalization of those two interfaces helps to understand the changes 😄

kyeah (approved):
LGTM!

PR #144, opened 2023-07-11
BCDA-7119: Add test coverage for main SSAS file

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7119

## 🛠 Changes

- Adds testing for the main.go file

## ℹ️ Context for reviewers

This file is the main entrypoint for the program and all related scripts. In order to provide better test coverage over these lines of code, I had to restructure the methods a bit to rely less on global package variables and move most of the main() function into separate, smaller functions that could take in a Flags object.

I also split the start() function so that we could test the "CreateServers" functionality without actually starting web servers. I think we could _probably_ write a test that actually covers starting the web servers and hitting them with HTTP requests but I had some trouble figuring that out, so I left that alone for now. This should still cover a good amount of the file.

## ✅ Acceptance Validation

Unit tests pass.

## 🔒 Security Implications

### DIFF
--- a/ssas/service/main/main.go
+++ b/ssas/service/main/main.go
@@ -152,31 +152,31 @@ func handleFlags(flags Flags) {
 		return
 	}
 	if flags.doStart {
-		start()
+		publicServer, adminServer, forwarder, err := createServers()
+
+		if err != nil {
+			os.Exit(-1)
+		}
+
+		start(publicServer, adminServer, forwarder)
 		return
 	}
 }
 
-func start() {
-	ssas.Logger.Infof("%s", "Starting ssas...")
-
+func createServers() (*service.Server, *service.Server, *http.Server, error) {
 	ps := public.Server()
 	if ps == nil {
 		ssas.Logger.Error("unable to create public server")
-		os.Exit(-1)
+		return nil, nil, nil, errors.New("unable to create public server")
 	}
 	ps.LogRoutes()
-	ps.Serve()
 
 	as := admin.Server()
 	if as == nil {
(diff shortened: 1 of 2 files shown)


ssas/service/main/main.go
```diff
@@ -52,86 +52,97 @@ import (
...
+	return flags
 }
 
-// We provide some simple commands for bootstrapping the system into place. Commands cannot be combined.
-func main() {
-	ssas.Logger.Info("Home of the System-to-System Authentication Service")
-
-	flag.Parse()
-	if doAddFixtureData {
+func handleFlags(flags Flags) {
```
laurenkrugen-navapbc, line 122: are these flags set from environment variables?

ssas/service/main/main.go
```diff
@@ -52,86 +52,97 @@ import (
...
-var doResetSecret bool
-var doNewAdminSystem bool
-var doListIPs bool
-var doListExpCreds bool
-var doShowXData bool
-var doStart bool
-var clientID string
-var auth string
-var systemName string
+type Flags struct {
```
laurenkrugen-navapbc, line 55: this cleans up the tests and the flags nicely. Thanks for adding!

laurenkrugen-navapbc (approved):
LGTM! thanks for the context in the PR description.

ssas/service/main/main.go
kyeah: They're passed in directly when the compiled go program is run, so something like `bcda --add-fixture-data` 😄

PR #148, opened 2023-11-22
BCDA-7416: Remove duplicate logging

karinamzalez:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7416

## 🛠 Changes

(What was added, updated, or removed in this PR.)
- removes duplicate logging in api.go 
- adds comments for future refactors that would improve logging

## ℹ️ Context for reviewers

(Background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers.)

- in service/main/main.go > listIPs() : replace panic with ssas.Logger.Fatal
- in server.go > NewServer > ln 164: add return statement after error is logged, so more generic log outside of this condition doesn't log error again. 
- in public/api.go: refactor error logging in ValidateSecret. 
- Adds three "TODOs" to bring attention to potential areas for refactor to streamline logging. TODOs are for the following:
   - Remove logging from JSONError and assert that 130 use cases of this function still have, more specific, logging outside of function 
   - Remove OperationFailed and refactor use cases to use ssas.Logger
   - Refactor ChooseSigningKey so error logging happens outside of function 

## ✅ Acceptance Validation

## 🔒 Security Implications

### DIFF
--- a/ssas/service/public/api.go
+++ b/ssas/service/public/api.go
@@ -249,7 +249,7 @@ func token(w http.ResponseWriter, r *http.Request) {
 		service.JSONError(w, http.StatusUnauthorized, http.StatusText(http.StatusUnauthorized), "invalid client id")
 		return
 	}
-	err = ValidateSecret(system, secret, w, r)
+	err = ValidateSecret(system, secret, r)
 	if err != nil {
 		ssas.Logger.Error("The client id and secret cannot be validated: ", err.Error())
 		return
@@ -292,24 +292,16 @@ func token(w http.ResponseWriter, r *http.Request) {
 	render.JSON(w, r, m)
 }
 
-func ValidateSecret(system ssas.System, secret string, w http.ResponseWriter, r *http.Request) (err error) {
+func ValidateSecret(system ssas.System, secret string, r *http.Request) (err error) {
 	savedSecret, err := system.GetSecret(r.Context())
-	if err != nil {
-		ssas.Logger.Errorf("Error getting secret: %s", err.Error())
-		service.JSONError(w, http.StatusUnauthorized, http.StatusText(http.StatusUnauthorized), "Error getting secret")
-		return err
-	} else if !ssas.Hash(savedSecret.Hash).IsHashOf(secret) {
-		ssas.Logger.Errorf("The incoming client secret is invalid")
-		service.JSONError(w, http.StatusUnauthorized, http.StatusText(http.StatusUnauthorized), constants.InvalidClientSecret)
-		return errors.New(constants.InvalidClientSecret)
+	if !ssas.Hash(savedSecret.Hash).IsHashOf(secret) {
+		err = errors.New(constants.InvalidClientSecret)
 	}
 
 	if savedSecret.IsExpired() {
-		ssas.Logger.Error("Credentials were expired")
-		service.JSONError(w, http.StatusUnauthorized, http.StatusText(http.StatusUnauthorized), "credentials expired")
-		return errors.New("The saved client secret is expired")
+		err = errors.New("The saved client credendials are expired")
 	}
-	return nil
+	return err
 }
 
 func tokenV2(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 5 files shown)


ssas/service/public/api.go
```diff
@@ -292,24 +292,16 @@ func token(w http.ResponseWriter, r *http.Request) {
 	render.JSON(w, r, m)
 }
 
-func ValidateSecret(system ssas.System, secret string, w http.ResponseWriter, r *http.Request) (err error) {
+func ValidateSecret(system ssas.System, secret string, r *http.Request) (err error) {
```
laurenkrugen-navapbc, line 295: We could remove the `r http.request` parameter if we are not using it here!

laurenkrugen-navapbc: we would also want to update the return type to include the http status code that gets returned depending on the `if` conditionals. So you could do something like

```
if condition {
    err = errors.New(...)
    status = http.<status>
```

ssas/service/public/api.go
```diff
@@ -249,7 +249,7 @@ func token(w http.ResponseWriter, r *http.Request) {
 		service.JSONError(w, http.StatusUnauthorized, http.StatusText(http.StatusUnauthorized), "invalid client id")
 		return
 	}
-	err = ValidateSecret(system, secret, w, r)
+	err = ValidateSecret(system, secret, r)
 	if err != nil {
 		ssas.Logger.Error("The client id and secret cannot be validated: ", err.Error())
```
laurenkrugen-navapbc, line 254: because we have removed the `service.JSONError(...)` in `ValidateSecret`, we should add `JsonError` to the line below the log statement here so we are still returning the error/http response.

ssas/service/server.go
```diff
@@ -160,6 +161,7 @@ func (s *Server) LogRoutes() {
 	routes, err := s.ListRoutes()
 	if err != nil {
```
laurenkrugen-navapbc, line 162: We could probably update this section to report an error back if there is one, instead of logging it as informational:
```
	if err != nil {
		ssas.Logger.Errorf("%s routing error: %v", banner, err)
	} 
```

We also don't need to return here, so we could remove that on L164!

ssas/service/server.go
```diff
@@ -281,6 +283,7 @@ func doHealthCheck(ctx context.Context) bool {
 	}
 
 	if err = db.Ping(); err != nil {
+		// (?) don't know if it actually gets to this error
 		ssas.Logger.Error("health check: database ping error: ", err.Error())
```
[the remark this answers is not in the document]
laurenkrugen-navapbc: In the err statement above this, err is the return value of `ssas.Connection.WithContext(ctx).DB()`; this is checking the return from `ssas.Connection.WithContext(ctx).DB()`. 

L285 is checking the error returned from db.Ping(), so both checks are valid and looking for errors returned by the function being called!

ssas/service/public/api.go
karinamzalez: @laurenkrugen-navapbc we are using `r http.request` on line 296. I considered returning the http status but it was exactly the same for each condition (http.statusunauthorized)

ssas/service/server.go
karinamzalez: @laurenkrugen-navapbc would ssas.logger.errorf exit out of the fucntion?

ssas/service/public/api.go
laurenkrugen-navapbc: this looks good! One thing to consider is that we are changing the return value of this call from one of three options [`Error getting secret`, `The incoming client secret is invalid`, or `credentials expired`] to `Error getting secret`.  

I had a quick look through bcda-app and we are only returning the http status to our consumers and we aren't using the error message returned from `validateSecret` elsewhere, so this change should be fine but we should keep an eye out, just in case.

ssas/service/server.go
laurenkrugen-navapbc: nope, it will just [log it at the level `error`](https://pkg.go.dev/github.com/sirupsen/logrus#Errorf)

laurenkrugen-navapbc:
This looks good overall, just need a couple things:

1. looks like tests are failing. are the unit-tests updated to reflect the changes?
2. can we also please update the PR title to be in the format of `BCDA-XXXX: <...>`?

PR #151, opened 2023-12-28
Karina/bcda 7415 ctx logger

karinamzalez:
## 🎫 Ticket

## 🛠 Changes

(What was added, updated, or removed in this PR.)

## ℹ️ Context for reviewers

(Background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers.)

## ✅ Acceptance Validation

## 🔒 Security Implications


ssas/service/public/api.go
```diff
@@ -69,7 +70,7 @@ func ResetSecret(w http.ResponseWriter, r *http.Request) {
 	setHeaders(w)
 
 	if rd, err = readRegData(r); err != nil || rd.GroupID == "" {
-		service.GetLogEntry(r).Println("missing or invalid GroupID")
+		log.GetCtxLogger(r.Context()).Println("missing or invalid GroupID")
```
laurenkrugen-navapbc, line 73: For these error handling sections that contain a log entry and `service.JSONError`, there will be two logs generated, since `JSONError` [also logs an event](https://github.com/CMSgov/bcda-ssas-app/blob/master/ssas/service/api_common.go#L34).

We could pass along a string into the JSONError function (missing or invalid, etc) + the ctx, or remove the log statement from the function and have the error or info logged out here in the error handling.

log/logger.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/CMSgov/bcda-ssas-app/ssas/constants"
+	"github.com/sirupsen/logrus"
+)
+
+// Logger provides a structured logger for this service
+var Logger logrus.FieldLogger
+
+// Event contains the superset of fields that may be included in Logger statements
+type Event struct {
```
laurenkrugen-navapbc, line 18: We have this `Event` struct and the `init()` for the logger defined in ssas/logger.go - can we remove one of the dupes? 😃

ssas/service/server.go
```diff
@@ -84,15 +81,15 @@ func ChooseSigningKey(signingKeyPath, signingKey string) (*rsa.PrivateKey, error
 
 // NewServer correctly initializes an instance of the Server type.
 func NewServer(name, port, version string, info interface{}, routes *chi.Mux, notSecure bool, useMTLS bool, signingKey *rsa.PrivateKey, ttl time.Duration, clientAssertAud string) *Server {
-
+	logger := log2.GetCtxLogger(context.Background())
```
laurenkrugen-navapbc, line 84: can we replace the [package](https://github.com/CMSgov/bcda-ssas-app/blob/95aad33ee6b1eb550bad7dc5a77c94939af2f78d/ssas/servi) `log` altogether in this file and just use `github.com/CMSgov/bcda-ssas-app/log` instead?

That way, we can rename `log2` and we are having an additional logger from the standard library also creating log events.

ssas/service/server.go
```diff
@@ -159,17 +156,19 @@ func (s *Server) ListRoutes() ([]string, error) {
...
+	logger.Infof("%s %v", banner, routes)
 }
 
 // Serve starts the server listening for and responding to requests.
 func (s *Server) Serve() {
+	logger := log2.GetCtxLogger(context.Background())
 	if s.notSecure {
-		ssas.Logger.Infof("starting %s server running UNSAFE http only mode; do not do this in production environments", s.name)
+		logger.Infof("starting %s server running UNSAFE http only mode; do not do this in production environments", s.name)
 		go func() { log.Fatal(s.server.ListenAndServe()) }()
```
laurenkrugen-navapbc, line 172: we can replace `log.Fatal` here with our context logger!

ssas/monitoring/monitoring.go
```diff
@@ -1,11 +1,12 @@
 package monitoring
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"os"
 
-	"github.com/CMSgov/bcda-ssas-app/ssas"
+	"github.com/CMSgov/bcda-app/log"
```
laurenkrugen-navapbc, line 9: This is an import for the bcda-app; should this be referencing another package in the SSAS-app [where the context logger](https://github.com/CMSgov/bcda-ssas-app/pull/151/files#diff-330686dc447b6523e8cbb02ae01e63ed53f3b214274ecc0da55c978965dda5ddR79) lives?

go.mod
```diff
@@ -19,27 +19,41 @@ require (
 )
 
 require (
+	github.com/CMSgov/bcda-app v0.0.0-20240116200610-366105d4b80b // indirect
```
laurenkrugen-navapbc, line 22: We should remove this dependency and update the go.mod and go.sum files, based on my comment [here](https://github.com/CMSgov/bcda-ssas-app/pull/151/files#r1467953507)!

laurenkrugen-navapbc:
I think there are a couple of things that might need to be updated:

1. when calling `log.GetCtxLogger(context.Background())`, `GetCtxLogger` is expecting context that will have a key value pair, specifically [one of the type `CtxLoggerKey`](https://github.com/CMSgov/bcda-ssas-app/pull/151/files#diff-330686dc447b6523e8cbb02ae01e63ed53f3b214274ecc0da55c978965dda5ddR80). I would expect that passing in a new `context.Background()` will result in nil reference errors, so we should probably update those to utilize the request context!
2. For some areas of the codebase where there is not a request context, such as startup functions like `init()`. It may be necessary to use the global logger for those places: `log.Logger`.

PR #166, opened 2024-03-21
BCDA-7415: Update SSAS to use ctx logger

laurenkrugen-navapbc:
## 🎫 Ticket

## 🛠 Changes

- TrackingIDs have been removed as args for some functions since the ID generated was not consistent and `transaction id` (uuid) is the agreed upon name convention  and type for correlating logs.
- `Op:<func>` has been kept in for consistency but open to updating
- `JSONError` no logger generates a log statement to eliminate duplicate logging where it is typically called
- `JSONError` status description uses const instead of free text
- a couple of tests were updated to use table driven structures
 
## ℹ️ Context for reviewers

Breaking these logging updates into smaller PRs. SSAS currently has a substantial amount of duplicate and context-less logs. In an effort to reduce errors being reported multiple times over and to make them more informative, this PR focuses on:
- surfacing errors at the top level function call
- logging errors at the top level function call
- removing some duplicate errors (this PR is already fairly large, so more removal will happen in another pr)
- reducing the usage of the `Operation<called|failed|etc>` and using the context logger so we have a standard of logging

I will have one more follow up PR to remove the  `Operation<called|failed|etc>` altogether from the remaining instances in the repo. I left that out to reduce an already large PR, but happy to include if reviewers request.

**There are several things that I will be ticketing for the upcoming sprint:**
- there are errors that occur at the ORM level, and instead of utilizing the built-in types, we do a catch-all statement and report an error that may be not be correct. Ex: "system not found" may be reported when the connection failed or transaction had to roll back, etc.
- we can reduce the number of free-text errors by introducing custom error types and also constants for holding error strings. there are use cases for both in the SSAS app.
- as a second part of the above bullet, we can return more informative error messages when they occur by check the error type (once we create some error types) for 4XX type responses or more specifically, 400 level responses. We should be returning which part of the request is invalid so consumers are not blocked until they make a GG post.

## ✅ Acceptance Validation

updated tests, ran locally, tests pass. Will be posting screenshots of local logs vs lower env log morning of 3/22.

## 🔒 Security Implications


ssas/groups.go
```diff
@@ -51,17 +51,28 @@ type GroupList struct {
 	Groups     []GroupSummary `json:"groups"`
 }
 
-func CreateGroup(ctx context.Context, gd GroupData, trackingID string) (Group, error) {
-	event := Event{Op: "CreateGroup", TrackingID: trackingID}
-	OperationStarted(event)
+type GroupData struct {
```
laurenkrugen-navapbc, line 54: no functional difference, moving the types to the top of the file for readability and per go's recommendation.

ssas/groups_test.go
```diff
@@ -65,7 +65,7 @@ func (s *GroupsTestSuite) TestCreateGroup() {
 	gd := GroupData{}
 	err := json.Unmarshal([]byte(fmt.Sprintf(SampleGroup, gid, SampleXdata)), &gd)
 	assert.Nil(s.T(), err)
-	g, err := CreateGroup(context.Background(), gd, RandomHexID())
```
laurenkrugen-navapbc, line 68: removal of tracking ID; not needed since we have transaction id.

ssas/systems_test.go
```diff
@@ -482,35 +495,37 @@ func (s *SystemsTestSuite) TestRegisterSystemIps() {
...
-		assert.NotEmpty(creds, address+"should have been a valid IP")
-		system, err := GetSystemByID(context.Background(), creds.SystemID)
-		assert.Nil(err)
-		ips, err := system.GetIPs()
-		assert.Nil(err)
-		assert.Equal([]string{address}, ips)
-		ips, err = GetAllIPs()
-		assert.Nil(err)
-		assert.Contains(ips, address)
+	for _, tc := range tests {
```
laurenkrugen-navapbc, line 498: updated test to use table driven structure

ssas/systems_test.go
```diff
@@ -457,20 +461,29 @@ func (s *SystemsTestSuite) TestRegisterSystemMissingData() {
 func (s *SystemsTestSuite) TestRegisterSystemIps() {
 	assert := s.Assert()
 
+	type test struct {
```
laurenkrugen-navapbc, line 464: updated to use table driven structure

ssas/service/admin/api.go
```diff
@@ -146,61 +142,71 @@ Responses:
...
-		service.JSONError(w, http.StatusNotFound, "", fmt.Sprintf("could not find system %s", id))
+		logger.Errorf("could not find system %s", id)
+		service.JSONError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound), fmt.Sprintf("could not find system %s", id))
 		return
 	}
 
-	ips, _ := s.GetIPsData(r.Context())
-	cts, _ := s.GetClientTokens(r.Context(), trackingID)
-	eks, _ := s.GetEncryptionKeys(r.Context(), trackingID)
+	ips, err := s.GetIPsData(r.Context())
```
alex-dzeda, line 192: good catch on the unhandled errors, wonder if some of these are the silent ones we've seen in periods of high load.

alex-dzeda:
Big fan of the references to the errors / error texts instead of the manual descriptions, the increased amount of error handling, and the improvements this will have for debugging tricky issues that pop up. 
One q: is it possible for the operation to be overwritten and one end up lost? Eg one command intends a certain operation, but causes two due to logic. That’s the only regression I’m worried about (which isn’t much, given how hard it was to debug the SSAS logs previously).

ssas/systems.go
```diff
@@ -698,21 +663,21 @@ func (system *System) GetIps(ctx context.Context, trackingID string) ([]IP, erro
...
 
-	regEvent := Event{Op: "DeleteIP", TrackingID: trackingID, Help: "calling from ssas.DeleteIP()"}
+	regEvent := Event{Op: "DeleteIP", Help: "calling from ssas.DeleteIP()"}
 	OperationStarted(regEvent)
 
 	// Find IP to delete
 	err = Connection.WithContext(ctx).First(&ip, "system_id = ? AND id = ?", system.ID, ipID).Error
 	if err != nil {
 		regEvent.Help = fmt.Sprintf("Unable to find IP address with ID %s: %s", ipID, err)
 		OperationFailed(regEvent)
```
alex-dzeda, line 679: Is there a different ticket that tracks the rest of the OperationFailed, OperationStarted, etc?

ssas/service/admin/api.go
```diff
@@ -336,57 +339,69 @@ Responses:
...
+		service.JSONError(w, http.StatusBadRequest, http.StatusText(http.StatusBadRequest), "")
 		return
 	}
 
-	ssas.OperationCalled(ssas.Event{Op: "RegisterClient", TrackingID: sys.TrackingID, Help: "calling from admin.createSystem()"})
+	logger.Infof("Operation Called: admin.createSystem()")
 	creds, err := ssas.RegisterSystem(r.Context(), sys.ClientName, sys.GroupID, sys.Scope, sys.PublicKey, sys.IPs, sys.TrackingID)
 	if err != nil {
-		service.JSONError(w, http.StatusBadRequest, fmt.Sprintf("could not create system; %s", err), "")
+		logger.Errorf("failed to create system; %s", err)
```
alex-dzeda, line 353: can we add a comment in here to come back to the error returned at a future time? It looks like this may not necessarily be a bad request, it may be a more temporary status.

alex-dzeda: Maybe there needs to be a ticket looking at the errors we return in SSAS overall; outside of the scope of this work, though

laurenkrugen-navapbc: yup, I actually have a placeholder ticket for this to address the return statuses because many of them are not helpful or do not fit the outcome of the error. https://jira.cms.gov/browse/BCDA-7954

I have also created another ticket to update some of our errors that we are returning from functions that make database calls, since those may be inaccurate: https://jira.cms.gov/browse/BCDA-7955

laurenkrugen-navapbc:
> Big fan of the references to the errors / error texts instead of the manual descriptions, the increased amount of error handling, and the improvements this will have for debugging tricky issues that pop up. One q: is it possible for the operation to be overwritten and one end up lost? Eg one command intends a certain operation, but causes two due to logic. That’s the only regression I’m worried about (which isn’t much, given how hard it was to debug the SSAS logs previously).

These shouldn't overwritten unless the line setting the log field is called again down the call stack.

alex-dzeda (approved):

PR #167, opened 2024-03-26
BCDA-7415: Remove unneeded logging

laurenkrugen-navapbc:
## 🎫 Ticket

## 🛠 Changes

- removing references to operation<called|failed|started|etc> type functions and event struct
- removing logging altogether in functions so logs are generated in the endpoint func or during startup processes

## ℹ️ Context for reviewers

Event logging functions did not provide useful context when called and were often called many times for the duration of a request, which generates unneeded log events. To keep the implementation consistent between SSAS and the api, the event type logging was removed and the context logger is now utilized where applicable.

## ✅ Acceptance Validation

tests pass and modified

## 🔒 Security Implications


ssas/systems.go
```diff
@@ -220,42 +219,27 @@ func (system *System) GetSecret(ctx context.Context) (Secret, error) {
...
 	}
 
 	err = system.deactivateSecrets(ctx)
 	if err != nil {
-		revokeCredentialsEvent.Help = "unable to revoke credentials for clientID " + system.ClientID
-		OperationFailed(revokeCredentialsEvent)
-		return fmt.Errorf("unable to revoke credentials for clientID %s: %s", system.ClientID, err.Error())
+		return fmt.Errorf("unable to revoke credentials for clientID " + system.ClientID)
 	}
-
```
alex-dzeda, line 256: Do we lose anything with the removal of this line? Only one where I don't see a log transferred along in the new version

alex-dzeda: I see it, upstream.

alex-dzeda (approved):

PR #174, opened 2024-04-22
BCDA-7212: Make 500 response errors properly reflect the error they encounter

austincanada:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7212

## 🛠 Changes

Updated error status on certain endpoint failures to properly describe what went wrong.

## ℹ️ Context for reviewers

Some 500 errors were being sent incorrectly for a Bad Requests.

## ✅ Acceptance Validation

## 🔒 Security Implications


alex-dzeda:
These look good, what are your thoughts on the others in the file that generally have an error of starting with "failed to" X and then have a BadRequest response? At least some of those appear to be transient / more appropriate as 5XX codes.

austincanada:
@alex-dzeda that's a good question. Some of them are okay as 500, such as marshaling responses, but unmarshaling should probably be a BadRequest.

alex-dzeda:
> @alex-dzeda that's a good question. Some of them are okay as 500, such as marshaling responses, but unmarshaling should probably be a BadRequest.

Can you look at those and see if you'd change any others? For example, looking at https://github.com/CMSgov/bcda-ssas-app/blob/7ed74be062d2d9cd7087483bcd4ba702488f6a06/ssas/service/admin/api.go#L52, it may not necessarily be a bad request, could be a transient DB error. I think looking over the CRUD operations (since that's where these tend to pop up) and updating those would be good with this ticket, since those listed in the ticket are just examples and not exhaustive.

ssas/service/admin/api.go
```diff
@@ -272,7 +272,7 @@ func updateSystem(w http.ResponseWriter, r *http.Request) {
 	_, err = ssas.UpdateSystem(r.Context(), id, v)
 	if err != nil {
 		logger.Errorf("failed to update system; %s", err)
-		service.JSONError(w, http.StatusBadRequest, http.StatusText(http.StatusBadRequest), "failed to update system")
+		service.JSONError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound), "failed to update system")
```
alex-dzeda, line 275: image
This error at least partially rolls up to the above image; I think it may be better to return the contextually-relevant http status, since this error appears to have multiple potential options for the correct status

alex-dzeda: eg from the above method alone, I see a case where a 400 or a 404 is appropriate, depending on the error

ssas/service/admin/api.go
```diff
@@ -308,7 +308,7 @@ func deleteGroup(w http.ResponseWriter, r *http.Request) {
 	err := ssas.DeleteGroup(r.Context(), id)
 	if err != nil {
 		logger.Errorf("failed to delete group; %s", err)
-		service.JSONError(w, http.StatusBadRequest, http.StatusText(http.StatusBadRequest), "failed to delete group")
+		service.JSONError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound), "failed to delete group")
```
alex-dzeda, line 311: Comment above applies here, too

ssas/service/admin/api.go
```diff
@@ -592,7 +598,7 @@ func registerIP(w http.ResponseWriter, r *http.Request) {
 	system, err := ssas.GetSystemByID(r.Context(), systemID)
 	if err != nil {
 		logger.Errorf("failed to retrieve system; %s", err)
-		service.JSONError(w, http.StatusNotFound, http.StatusText(http.StatusBadRequest), "Invalid system ID")
+		service.JSONError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound), "Invalid system ID")
```
alex-dzeda, line 601: Comment above here

ssas/service/admin/api.go
austincanada: Had a conversation with Lauren. Rolling new conditional logic for custom response handling based on downstream errors have involved extra scope that is outside this specific ticket. Will create a new ticket for the custom error handling outside of malformed requests.

laurenkrugen-navapbc (approved):

PR #187, opened 2024-09-20
Fix calls to fmt Errorf with no interpolation

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8373

## 🛠 Changes

Convert various fmt.Errorf statements to either require the interpolation or to errors.New.  Both versions should return a string as an error.

## ℹ️ Context

The build was failing on golangci-lint checks.

## 🧪 Validation

Linting and testing were run to verify no breaking changes.


carlpartridge:
> LGTM. Does it need any test updates?

Theoretically no, the functionality should be unchanged.

PR #197, opened 2024-12-30
BCDA-8360 add waf sync lambda

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8360

## 🛠 Changes

Add WAF auto sync lambda

## ℹ️ Context

We have recently switched over to WAF to control our allowlist of IP addresses.  This lambda should sync our app DB with the source of truth on IP addresses to our WAF.

## 🧪 Validation

Local tests and linting passing.


lambda/wafsync/aws.go
```diff
@@ -0,0 +1,93 @@
...
+	Scope     string
+	LockToken string
+	Addresses []string
+}
+
+var createSession = func() (*session.Session, error) {
+	sess := session.Must(session.NewSession())
+
+	var err error
+	if isTesting {
```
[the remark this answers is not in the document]
carlpartridge: It is declared in main.go (`var isTesting = os.Getenv("IS_TESTING") == "true"`).  It seems to be used in a few places throughout the code to set things in certain ways.

lambda/wafsync/db.go
```diff
@@ -0,0 +1,91 @@
...
+	for rows.Next() {
+		var ip net.IP
+
+		err = rows.Scan(&ip)
+		if err != nil {
+			log.Warningf("Scan error: %+v", err)
+			return nil, err
+		}
+
+		// Not sure why we need this
```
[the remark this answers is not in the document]
carlpartridge: Yeah as far as I can tell the only thing its doing is logging the number of rows read.  Im thinking this could be useful in the early stages to debug any issues?

lambda/wafsync/db.go
```diff
@@ -0,0 +1,91 @@
...
+	Exec(context.Context, string, ...interface{}) (pgconn.CommandTag, error)
+	QueryRow(context.Context, string, ...interface{}) pgx.Row
+	Query(context.Context, string, ...interface{}) (pgx.Rows, error)
+	Ping(context.Context) error
+	Prepare(context.Context, string, string) (*pgconn.StatementDescription, error)
+	Close(context.Context) error
+}
+
+func getValidIPAddresses(ctx context.Context, conn PgxConnection) ([]string, error) {
+	defer conn.Close(ctx)
```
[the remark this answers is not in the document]
carlpartridge: My understanding is that defers are last in first out, so if we want this to happen at the very end we want to defer it sooner rather than later.  I dont think it matters too much but I think it should at least come before the `defer rows.Close()` on line 52.

lambda/wafsync/README.md
```diff
@@ -0,0 +1,12 @@
+Get short-term AWS credentials through CloudTamer:
+You can run from the base dir (bcda-ssas-app) using the following command:
+- `make test-path TEST_PATH="lambda/wafsync/*.go"`.  (You might have to `make load-fixtures` first).
```
laurenkrugen-navapbc, line 3: I think we should run this manually (outside of GA) by doing it through AWS console vs `make`. Seems like this could be run by accident and keeps sensitive data off our machines if we are troubleshooting. make load-fixtures can also be blocked if we have exceeded our github public token usage, which would prevent us from running this purposefully. Thoughts?

lambda/wafsync/db.go
```diff
@@ -0,0 +1,91 @@
...
+	count := 0
+	ipAddresses := []string{}
+	defer rows.Close()
+
+	for rows.Next() {
+		var ip net.IP
+
+		err = rows.Scan(&ip)
+		if err != nil {
+			log.Warningf("Scan error: %+v", err)
```
laurenkrugen-navapbc, line 60: What do you think about bumping this to a fatal or error level? if scan fails on reading any of the IP set that should be updated, then that would mean that the IP allow list could be out of sync between database <> waf.

.github/workflows/waf-sync-lambda-integration-test.yml
```diff
@@ -0,0 +1,46 @@
...
+      - uses: actions/checkout@v4
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-dev-github-actions
+      - name: Send event to trigger lambda function
+        id: invoke-lambda
+        run: |
+          echo "STARTTIME=`date +%s`" >> "$GITHUB_OUTPUT"
+          aws lambda invoke --function-name bcda-dev-api-waf-sync test-result.txt
```
laurenkrugen-navapbc, line 46: 1. if an error is returned; do we need to do any verification here or do we expect a non-zero exit code?
2.  do we want to add any additional validation for other cases where it runs successfully, but the end result might be different from what we expect; ie the update doesn't return an error, but there are no IPs in the WAF?

carlpartridge: 1. I believe returning nonzero is enough but Im not entirely sure how to force the function to error out.
2. I feel like that would best be solved via splunk or cloudwatch alerts?  They could watch for a variety of things based off the logs?

lambda/wafsync/db.go
carlpartridge: > What do you think about bumping this to a fatal or error level? if scan fails on reading any of the IP set that should be updated, then that would mean that the IP allow list could be out of sync between database <> waf.

I think error makes more sense as we already return nil, err on issue here.  Also I think returning fatal is not recommended in lambdas.

.github/workflows/waf-sync-lambda-integration-test.yml
carlpartridge: > 1. [It looks like it should return a status code and function error](https://docs.aws.amazon.com/lambda/latest/api/API_Invoke.html#API_Invoke_ResponseSyntax) (among a few other fields), if any occurred.
> 
> It looks like each time the dev lambda is invoked, an error is being returned at the moment, but the integration test says that it's passing: `Phase: invoke Status: error Error Type: Runtime.ExitError`
> 
> Could we update this so that we look for a status code and a function error, since both could indicate a problem?
> 
> 2. We could add a cli command to pull a list from the waf and if it's length is zero, then we could fail?

Its interesting the `aws lambda invoke ...` returns a json response as mentioned, however the output file does not include that information.  Best I could come up with is testing that we are not seeing an error in the output file.

I think checking the waf for issues is best done via cloudwatch (for your specific example) as well as checking splunk logs more generally.

laurenkrugen-navapbc: yup, I think we could setup the alerting around cloudwatch for sure.

laurenkrugen-navapbc (approved):
Approving, with the steps I took to validate:

- review lambda invocation to ensure that it's running with no errors (it is)
- review logs for success messages (they are present)

The only thing that we should do before we cutover is put some test data in our lower env to make sure it works properly. I looked in the dev WAF, and there are no IP addresses. Maybe we can coordinate with @gfreeman-navapbc on some safe-for-use IPs that we can dump in there in the interim?

PR #203, opened 2025-03-05
BCDA-8634: Add tag and release workflow

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8634

## 🛠 Changes

Add tag and release workflow

## ℹ️ Context

Part of moving off jenkins

## 🧪 Validation

Local testing and linting


laurenkrugen-navapbc (approved):
Same comment as other api pr about manually typing out the current tag; is that still needed? Either way, LGTM!

PR #218, opened 2025-06-16
BCDA-9092: log cms ID when credentials are reset

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9092

## 🛠 Changes

Added CMS identifier in logs when:
- credentials are reset 
- credentials are revoked
- a new system is created

Updated tests for all of the above to ensure the CMS ID is present.

## ℹ️ Context

Added for better visibility around requests to rotate or create credentials.

## 🧪 Validation

Tests updated and passing.

### DIFF
--- a/ssas/service/admin/api.go
+++ b/ssas/service/admin/api.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -442,14 +444,23 @@ func resetCredentials(w http.ResponseWriter, r *http.Request) {
 
 	logger.Infof("Operation Called: admin.resetCredentials()")
 	creds, err := system.ResetSecret(r.Context())
+
 	if err != nil {
 		logger.Errorf("failed to reset secret: %s", err)
 		service.JSONError(w, http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError), "")
 		return
 	}
 
-	// Used for alerting; update alert if this line changes
-	logger.Infof("secret reset in group %s with XData: %s", system.GroupID, creds.XData)
+	g, err := ssas.GetGroupByID(r.Context(), strconv.FormatUint(uint64(system.GID), 10))
+	if err != nil {
+		logger.Error()
+		service.JSONError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound), "failed to get group")
+	}
+
+	exp := regexp.MustCompile(`(?:\[")(\w+)(?:"\])`)
+	cms_id := exp.FindStringSubmatch(g.XData)
+
+	logger.Infof("secret reset in group %s with XData: %s", cms_id[1], creds.XData)
 
 	credsJSON, err := json.Marshal(creds)
 	if err != nil {
--- a/ssas/service/admin/api.go
+++ b/ssas/service/admin/api.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"regexp"
-	"strconv"
 	"strings"
 	"time"
 
@@ -357,8 +355,15 @@ func createSystem(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	group, err := ssas.GetGroupByGroupID(r.Context(), sys.GroupID)
+	if err != nil {
+		logger.Errorf("could not get group XData for clientID %s: %s", creds.ClientID, err.Error())
+		service.JSONError(w, http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError), "")
+		return
+	}
+
 	// Used for alerting; update alert if this line changes
-	logger.Infof("system registered in group with XData: %s", creds.XData)
+	logger.Infof("system registered in group %s with XData: %s", group.GroupID, group.XData)
 
 	credsJSON, err := json.Marshal(creds)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


ssas/service/admin/api.go
```diff
@@ -442,14 +444,23 @@ func resetCredentials(w http.ResponseWriter, r *http.Request) {
...
+	g, err := ssas.GetGroupByID(r.Context(), strconv.FormatUint(uint64(system.GID), 10))
+	if err != nil {
+		logger.Error()
+		service.JSONError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound), "failed to get group")
+	}
+
+	exp := regexp.MustCompile(`(?:\[")(\w+)(?:"\])`)
+	cms_id := exp.FindStringSubmatch(g.XData)
+
+	logger.Infof("secret reset in group %s with XData: %s", cms_id[1], creds.XData)
```
carlpartridge, line 463: I would think the regex would grab from x_data (eg {"cms_id":["A9999"]}) just the array of CMS IDs (eg ["A9999"])?  If so then dont we want cms_ids[0]?  This was also previously spitting out the group id (which doesnt seem that useful) and the full x_data.  Is there any chance there would be anything sensitive in x_data?  Do we just want to spit out the CMS ID?

michaeljvaldes, line 463: If it's not too much trouble, it might be useful to create a unit test or add a condition to an existing test that checks for the presence of this message in the logs.
Since our alert is dependent on it, it could be useful to have a check to make sure it's there.

ssas/service/admin/api.go
```diff
@@ -442,14 +444,23 @@ func resetCredentials(w http.ResponseWriter, r *http.Request) {
...
-	// Used for alerting; update alert if this line changes
-	logger.Infof("secret reset in group %s with XData: %s", system.GroupID, creds.XData)
+	g, err := ssas.GetGroupByID(r.Context(), strconv.FormatUint(uint64(system.GID), 10))
+	if err != nil {
+		logger.Error()
+		service.JSONError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound), "failed to get group")
+	}
+
+	exp := regexp.MustCompile(`(?:\[")(\w+)(?:"\])`)
+	cms_id := exp.FindStringSubmatch(g.XData)
```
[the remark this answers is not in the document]
laurenkrugen-navapbc: I was also wondering this because at one point, I though we did have this ID being logged. Going to have another look through the code because making a second call to the db feels  a little clunky.

ssas/service/admin/api.go
laurenkrugen-navapbc: yup, the two tests will check for the presence of the [cms id](https://github.com/CMSgov/bcda-ssas-app/pull/218/files#diff-759c784a61aca45d4225a3cee3907f1e0961aeab2dd02663c1a34a44083b546fR475) in the logs!

laurenkrugen-navapbc: Yes, if we have more than one CMS ID then we would only grab the first one, but I can update to report all in the group. We don't have a way to correlate a single system back to a single CMS ID, which is not currently a problem because _most_ acos have a 1:1 correlation with a group. This could be a future problem though? If a group contains many CMS IDs, we would need to go back into the database to determine the system and then go from there.

ssas/service/admin/api.go
```diff
@@ -442,14 +444,23 @@ func resetCredentials(w http.ResponseWriter, r *http.Request) {
...
+
```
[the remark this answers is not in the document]
laurenkrugen-navapbc: all have been updated!

ssas/service/admin/api.go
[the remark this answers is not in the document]
laurenkrugen-navapbc: removed this entirely -- our logs apply regex, so only the x_data is needed and the logging alert will do the rest!

ssas/service/admin/api.go
laurenkrugen-navapbc: This will pull all of the CMS IDs and the logging alert will apply regex that should grab the IDs, although we should test that.

ssas/service/admin/api.go
```diff
@@ -355,8 +355,15 @@ func createSystem(w http.ResponseWriter, r *http.Request) {
...
+	group, err := ssas.GetGroupByGroupID(r.Context(), sys.GroupID)
+	if err != nil {
+		logger.Errorf("could not get group XData for clientID %s: %s", creds.ClientID, err.Error())
+		service.JSONError(w, http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError), "")
+		return
+	}
+
 	// Used for alerting; update alert if this line changes
-	logger.Infof("system registered in group with XData: %s", creds.XData)
+	logger.Infof("system registered in group %s with XData: %s", group.GroupID, group.XData)
```
carlpartridge, line 366: I think above these logger lines we still need to set the field into the logger, eg: `logger.WithFields(logrus.Fields{"cms_id": cms_id})`.  This is how the following gets set into the splunk logs themselves:
Screenshot 2025-06-18 at 4 40 30 PM
We can also see in the splunk alerts action that we reference the field cms_id (and result_uri):
Screenshot 2025-06-18 at 4 42 30 PM

laurenkrugen-navapbc: it doesn't pull it as a field in the log message itself, but it does put the regex result in a variable:

`match(msg, "secret reset in group") | rex "(?<cms_id>[ADEVCK]\d{3,4})"`

Because the CMS ID will be in the xdata, it will retrieve it via regex and refer to it by the variable name set above.

carlpartridge: Where is it getting uri from then?  Maybe it does both?  When it came to fixing the `BCDA Prod - ACO API Bulk Data Request` alert just getting the cms_id field into splunk fixed the issue.  I think its also potentially useful to have it as a field in splunk so as to query it better instead of pulling it out via regex.  But if you think this will work and you think the alternative is a lot more work then Im happy to approve this.

carlpartridge (approved):

PR #242, opened 2025-08-19
Carl bcda 9359 refactor docker image

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9359

## 🛠 Changes

Clean up of docker images.  Removed unnecessary packages/statements, updated for best practices, and added non-root user.

## ℹ️ Context

In preparation for fargate we want to clean up, update, and secure our docker images.

## 🧪 Validation

Local builds.

### DIFF
--- a/Dockerfiles/Dockerfile.package
+++ b/Dockerfiles/Dockerfile.package
@@ -1,16 +1,24 @@
 FROM golang:1.24.4-alpine3.22
 
+RUN apk update upgrade && apk add \
+    build-base \
+    gcc \
+    git \
+    gpg \
+    gpg-agent \
+    # ruby \
+    # ruby-dev \
+    rpm
+
 ENV CGO_ENABLED=0
 
-ARG GPG_PUB_KEY_FILE
-ARG GPG_SEC_KEY_FILE
+ARG GPG_PUB_FILE_PATH
+ARG GPG_SEC_FILE_PATH
 ARG GPG_RPM_USER
 ARG GPG_RPM_EMAIL
 ARG BCDA_GPG_RPM_PASSPHRASE
 
-RUN apk update upgrade
-RUN apk add build-base gcc ruby ruby-dev rpm git gpg gpg-agent
-RUN gem install --no-document fpm etc
+# RUN gem install --no-document fpm etc
 
 WORKDIR /go/src/github.com/CMSgov/bcda-ssas-app
 COPY . .
@@ -21,3 +29,6 @@ RUN chmod u+x build_and_package.sh
 
 ENTRYPOINT ["sh", "build_and_package.sh"]
 CMD []
+
+RUN addgroup -S ssas-group && adduser -S ssas-user -G ssas-group
+USER ssas-user
--- a/Dockerfiles/Dockerfile.ssas
+++ b/Dockerfiles/Dockerfile.ssas
@@ -1,9 +1,14 @@
+### OpenAPI docs
 FROM golang:1.24.4-alpine3.22 AS documentation
-RUN apk update upgrade
-RUN apk add git build-base gcc binutils-gold
+
+RUN apk update upgrade && apk add \
+    binutils-gold \
+    build-base \
+    gcc \
+    git
+
 WORKDIR /go/src/github.com/CMSgov/bcda-ssas-app
 COPY . .
-ARG GO111MODULE=auto
 
 RUN dir=$(mktemp -d) && \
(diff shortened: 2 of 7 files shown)


Dockerfiles/Dockerfile.package
```diff
@@ -21,3 +29,6 @@ RUN chmod u+x build_and_package.sh
 
 ENTRYPOINT ["sh", "build_and_package.sh"]
 CMD []
+
+RUN addgroup -S ssas-group && adduser -S ssas-user -G ssas-group
+USER ssas-user
```
carlpartridge, line 34: Not entirely sure this is useful?  As it comes after the entrypoint/cmd?  My general understanding of the security implications of using non-root user is that it prevents someone that is able to get onto a container from doing anything as root user.  So even if this comes at the very end it would still be useful?  Cant do this earlier as the script requires root level access.

Dockerfiles/Dockerfile.ssas
```diff
@@ -13,26 +18,48 @@ RUN dir=$(mktemp -d) && \
...
 WORKDIR /go/src/github.com/CMSgov/bcda-ssas-app/ssas
 COPY --from=documentation /go/src/github.com/CMSgov/bcda-ssas-app/ssas/swaggerui ./swaggerui
+
 RUN go build -ldflags "-X github.com/CMSgov/bcda-ssas-app/ssas/constants.Version=$VERSION" -o ssas ./service/main
 
+RUN addgroup -S ssas-group && adduser -S ssas-user -G ssas-group
+USER ssas-user
+
+
+### Final image
```
carlpartridge, line 44: We build/compile the ssas app in the previous stage (builder) then copy it over in this stage.  Was curious if there was a reason for this approach instead of starting this final stage from the builder image?  We could probably move some of this final stages work up to the builder stage and leave the final stage a bit smaller?

Dockerfiles/Dockerfile.package
laurenkrugen-navapbc: Yeah, not sure how useful this is at the end. 

It would prevent someone from using the root account if they managed to get access to the running container, so there's that?

laurenkrugen-navapbc (approved):

Dockerfiles/Dockerfile.tests
```diff
@@ -1,12 +1,18 @@
...
+RUN apk update upgrade && apk add \
+    bash \
+    build-base \
+    curl
 
-RUN apk add bash build-base curl
+# this image currently gets used by unit_test.sh which needs root level access to run various commands
+# unit_test.sh could likely get refactored away but for now leaving as is
+# RUN addgroup -S ssas-group && adduser -S ssas-user -G ssas-group
+# USER ssas-user
```
carlpartridge, line 11: Spent a while trying to get permissions working for the unit_test.sh script.  I think generally speaking there must be a better way to go about this than a shell script.  It mostly just does some setup (makes dirs, sets some env vars), then runs a few commands (go test sum, go tool cover, moves the test results, then drops the test db).  We want to be able to run this locally as well as part of CI.  Feels like there is a better way to approach this and a new ticket is probably for the best here.  I feel like leaving this comment in as a guide as to why we are currently not swapping over to non-root user.

Dockerfiles/Dockerfile.ssas
carlpartridge: Thinking out the loud:  The final image could probably just pull from alpine?  Is that worth looking into?

The build/compilation layer currently is basically the exact same as the final image layer, +/- one or two apk packages.  Just wondering why have a separation at all?  Another approach could be to move all apk packages into the build/compilation then re-use that for the final image instead of just copying certain things over.

Feels like theres a lot of options here and not sure if any are strictly better than others.

michaeljvaldes: For what it's worth, I think it makes a lot of sense as it is right now, as it's following some best practices:
1. the final image is small -- we're only copying what we absolutely need from the builder; e.g. the builder has git but the final image doesn't, we're also only copying over the ssas directories we need
2. I don't think there's anything else we can move from the final stage to the builder as the final stage is doing the bare minimum -- installing the final packages we need, copying over files, and setting a user (which I don't think can be moved to an earlier step)

carlpartridge: Ok yeah I think I generally agree with you but am not totally sold how much we are actually saving in this specific case.  Overall a fairly minor issue.  Appreciate the conversation!

laurenkrugen-navapbc (approved):

michaeljvaldes (approved):

PR #249, opened 2025-09-26
BCDA-9481: remove systems.go init

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9481

## 🛠 Changes

- Init removed from systems.go and hash.go
- config.go added as a way to load environment variables that don't get injected via a dependency
- LoadEnvConfigs called in `main()`

## ℹ️ Context

Changes were made as larger initiative to remove init in the codebase. Init has been a development slowdown and blocker when troubleshooting singular tests.

## 🧪 Validation

Tests passing.


ssas/cfg/config.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type SystemsConfig struct {
+	DefaultScope         string
+	MaxIPs               int
+	CredentialExpiration time.Duration
+	MacaroonExpiration   time.Duration
+}
+
+func loadSystemConfig() *SystemsConfig {
```
laurenkrugen-navapbc, line 30: took the init func body and moved it here for the system env vars.

PR #252, opened 2025-10-04
BCDA-9493: Update jwt library

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9493

## 🛠 Changes

Updated our jwt library to new major version.

## ℹ️ Context

Older version had a medium vulnerability.

## 🧪 Validation

Local testing and linting.


ssas/service/public/api.go
```diff
@@ -358,22 +359,27 @@ func (h *publicHandler) tokenV2(w http.ResponseWriter, r *http.Request) {
...
+	tokenDuration := claims.ExpiresAt.Sub(claims.IssuedAt.Time)
+	if tokenDuration.Seconds() > 300 { // 5 minute max duration
 		service.JSONError(w, http.StatusBadRequest, "IssuedAt (iat) and ExpiresAt (exp) claims are more than 5 minutes apart", "")
 		return
 	}
 
+	if tokenDuration.Seconds() < 0 { // issued at AFTER expires at
+		service.JSONError(w, http.StatusBadRequest, "IssuedAt (iat) claim marked as AFTER ExpiresAt (exp) claim", "")
+		return
+	}
```
carlpartridge, line 381: This new check verifies that a token does not have its IssuedAt set to after its ExpiresAt is set to.  Makes sense logically and was already being tested for in our test suite.  I guess the changes to how `tokenDuration` was calculated for this check required a new check instead of the above check.

ssas/service/public/tokens.go
```diff
@@ -98,22 +98,21 @@ func tokenValidity(ctx context.Context, tokenString string, requiredTokenType st
 		logger.Error(err)
 		return err
 	}
-
-	c := t.Claims.(*service.CommonClaims)
-
-	err = checkAllClaims(c, requiredTokenType)
-	if err != nil {
+	if !t.Valid {
```
michaeljvaldes, line 101: As a bonus, this function is more readable now 👍

michaeljvaldes (approved):

PR #253, opened 2025-10-07
BCDA-9505: remove tokens.go init

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9505

## 🛠 Changes

remove init in tokens.go; initializing needed variables in config package. Moved init to function `SetAccessTokenCreator` that is called in main. 

## ℹ️ Context

Changes were made as larger initiative to remove init in the codebase. Init has been a development slowdown and blocker when troubleshooting singular tests.

## 🧪 Validation

Tests passing.

### DIFF
--- a/ssas/cfg/config.go
+++ b/ssas/cfg/config.go
@@ -10,13 +10,14 @@ import (
 )
 
 var (
-	DefaultScope         string
-	MaxIPs               int
-	CredentialExpiration time.Duration
-	MacaroonExpiration   time.Duration
-	HashIter             int
-	HashKeyLen           int
-	SaltSize             int
+	DefaultScope                  string
+	MaxIPs                        int
+	CredentialExpiration          time.Duration
+	MacaroonExpiration            time.Duration
+	HashIter                      int
+	HashKeyLen                    int
+	SaltSize                      int
+	SelfRegistrationTokenDuration time.Duration
 )
 
 // Get configuration/environment variables for Hashes and Systems.
@@ -67,4 +68,7 @@ func LoadEnvConfigs() {
 		// ServiceHalted(Event{Help:"SSAS_HASH_ITERATIONS, SSAS_HASH_KEY_LENGTH and SSAS_HASH_SALT_SIZE environment values must be set"})
 		panic("SSAS_HASH_ITERATIONS, SSAS_HASH_KEY_LENGTH and SSAS_HASH_SALT_SIZE environment values must be set")
 	}
+
+	minutes := GetEnvInt("SSAS_MFA_TOKEN_TIMEOUT_MINUTES", 60)
+	SelfRegistrationTokenDuration = time.Duration(int64(time.Minute) * int64(minutes))
 }
--- a/ssas/service/public/tokens.go
+++ b/ssas/service/public/tokens.go
@@ -3,7 +3,6 @@ package public
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/CMSgov/bcda-ssas-app/ssas"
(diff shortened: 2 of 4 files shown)


ssas/cfg/config.go
```diff
@@ -67,4 +68,7 @@ func LoadEnvConfigs() {
 		// ServiceHalted(Event{Help:"SSAS_HASH_ITERATIONS, SSAS_HASH_KEY_LENGTH and SSAS_HASH_SALT_SIZE environment values must be set"})
 		panic("SSAS_HASH_ITERATIONS, SSAS_HASH_KEY_LENGTH and SSAS_HASH_SALT_SIZE environment values must be set")
 	}
+
+	minutes := GetEnvInt("SSAS_MFA_TOKEN_TIMEOUT_MINUTES", 60)
```
carlpartridge, line 72: Would it be useful to have a comment describing what we are setting here or do you think the key `SSAS_MFA_TOKEN_TIMEOUT_MINUTES` is good enough?

ssas/service/public/tokens.go
```diff
@@ -13,12 +12,7 @@ import (
 
 var accessTokenCreator TokenCreator
 
-var selfRegistrationTokenDuration time.Duration
-
-func init() {
-	minutes := cfg.GetEnvInt("SSAS_MFA_TOKEN_TIMEOUT_MINUTES", 60)
-	selfRegistrationTokenDuration = time.Duration(int64(time.Minute) * int64(minutes))
-
+func SetAccessTokenCreator() {
```
carlpartridge, line 15: All we seem to be doing is setting an empty `accessTokenCreator`?  And the only use of accessTokenCreator is that we call `GenerateToken()` on it?  Why have it at all?  It kind of seems like its a shim in place in case we wanted to change our implementation?

laurenkrugen-navapbc: Agreed - I spent some time on updating the implementation to remove the interface and some of these getter/setter methods but ended up reverting, because of the amount of changes to the test suite and wasn't sure if we wanted to take on that risk? If you're open to a fair amount of changes to the token tests, I can try to put those changes back in and we can discuss?

It's only in place because:

> // TokenCreator provides methods for the creation of tokens.
> // Currently only AccessTokenCreator implements this interface.
> // TO DO:
> // Define a MFATokenCreator & a RegistrationTokenCreator that will implement TokenCreator interface,
> // then add CreateCommonClaims to this interface that all 3 can share.

I ca

ssas/cfg/config.go
laurenkrugen-navapbc: I think that one is self explanatory!

ssas/service/public/tokens.go
```diff
@@ -54,7 +44,11 @@ func MintRegistrationToken(oktaID string, groupIDs []string) (*jwt.Token, string
 		return nil, "", err
 	}
 
-	return server.MintTokenWithDuration(&claims, selfRegistrationTokenDuration)
+	return server.MintTokenWithDuration(&claims, cfg.SelfRegistrationTokenDuration)
```
michaeljvaldes, line 47: Do we think we should have a default duration in case the duration isn't set in the config? It looks like it was 60 minutes before.

michaeljvaldes: Additionally, if we think the efficiency of checking the env var value each time is a concern, we can also maybe store the default duration in the state of the `TokenCreator` and just read the env var once upon creation

michaeljvaldes: Oops, nevermind -- I misunderstood how cfg works in SSAS. Please disregard!

PR #291, opened 2026-02-27
BCDA-9853: improve docker config and makefile commands

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9853
[related bcda-app PR](https://github.com/CMSgov/bcda-app/pull/1342)

## 🛠 Changes

- suppress noisy, expected error messages from migrations unit tests
- create a new makefile target `reset-db` which properly resets the database before applying fixtures
- reduce warnings raised when running docker commands
- introduce more healthchecks and a script that waits for a service to be healthy (replacement for sleep/wait-for-it) 
- reorganize some docker-related files into docker directory
- split the integration test docker services into their own compose file, which includes the main compose file
- rename compose files
- remove migration compose file in favor of a `docker run {image}` command
- enable `compose bake` in our main docker build command, reducing build time from ~4:40 to ~4:00

## ℹ️ Context

While building SSAS, there are a lot of expected errors from tests and docker warnings logged to the shell window that can be suppressed. 
Additionally, we are now using `sleep` to wait for service to be ready, and a more robust solution that waits for docker services to be healthy using their healthchecks is a useful upgrade.
Finally, collecting more docker-related files into a `/docker` directory organizes the repo just a bit more, and splitting the integration tests into their own compose file organizes them by the docker ecosystem in which they operate (the main/app one, which is "included" in this new compose file) and allows them to use the main services as dependencies. 

## 🧪 Validation

run `make docker-bootstrap` and `make test`


Makefile
```diff
@@ -38,29 +38,41 @@ setup-tests:
 test-path: setup-tests
 	@docker compose -f docker-compose.test.yml run --rm tests go test -v $(TEST_PATH)
 
+reset-db:
+	# Rebuild the databases to ensure that we're starting in a fresh state
+	docker compose rm -fsv db
+
+	docker compose up -d db
```
carlpartridge, line 45: How do you feel about `--remove-orphans` on these compose up commands?

Makefile
```diff
@@ -38,29 +38,41 @@ setup-tests:
...
-	docker compose -f docker-compose.migrate.yml run --rm migrate  -database "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -path /go/src/github.com/CMSgov/bcda-ssas-app/db/migrations up
-	docker compose -f docker-compose.yml run ssas sh -c 'ssas --add-fixture-data'
+	$(MAKE) reset-db
+	docker compose -f docker-compose.yml run --rm ssas sh -c 'ssas --add-fixture-data'
 
 docker-build:
-	docker compose build
-	docker compose -f docker-compose.test.yml build
+	docker compose build --force-rm
+	docker compose -f docker-compose.test.yml build --force-rm
```
carlpartridge: You probably understand better than I do but as far as I have experienced the two test docker compose file requires at least some of the services running in the default docker compose file.  Or at least the full test suite does?  Should we have these services shared across the two?  Is there a good way to do that?  Or is the current set up fine?

docker-compose.migrate.yml
```diff
@@ -1,5 +1,3 @@
-version: '3'
```
carlpartridge, line 1: I think with the changes to load-fixtures we can get rid of this file altogether?

docker-compose.yml
```diff
@@ -55,5 +56,12 @@ services:
       - "3103:3003"
       - "3104:3004"
       - "3105:3005"
+    healthcheck:
+      test: ["CMD", "curl", "-k", "-s", "-f", "-o", "/dev/null", "localhost:3003/_health"]
+      interval: 1m30s
```
bhagatparwinder: We wait for 60 seconds by default [here](https://github.com/CMSgov/bcda-ssas-app/pull/291/changes#diff-8c44dd1f82cc834d6fee7db94b7936c2c9c79f1787cd4bd85f8d9fe73b358d19R18) but compose file waits for 90 seconds. If `db/ssas` don’t report healthy on the first healthcheck run wouldn't the script time out even though the service will be healthy _soon_

Makefile
```diff
@@ -38,29 +38,41 @@ setup-tests:
...
 
-docker-bootstrap: docker-build load-fixtures
+docker-bootstrap:
+	$(MAKE) docker-build
+	docker compose up -d
```
bhagatparwinder: Do I understand this correctly? `docker-bootstrap` starts `ssas`, waits for it to become healthy, then calls `load-fixtures`, which internally runs `reset-db` and recreates the `db` service.

If this is the case, wouldn't resetting the DB underneath a running SSAS can flip SSAS health to unhealthy and _potentially_ keep it that way?

Makefile
michaeljvaldes: I actually think I got it to a place where the dangling `compose run` containers were all dealt with, but I added `--remove-orphans` to the main `compose up` command in the `docker-bootstrap` recipe just in case.

docker-compose.migrate.yml
michaeljvaldes: I think we still want the migrate compose file. We use it in the `reset-db` recipe (which is called in `load-fixtures`) to build the schema, after which we can load seed data.

Makefile
```diff
@@ -38,29 +38,41 @@ setup-tests:
...
-	docker compose build
-	docker compose -f docker-compose.test.yml build
+	docker compose build --force-rm
+	docker compose -f docker-compose.test.yml build --force-rm
 
-docker-bootstrap: docker-build load-fixtures
+docker-bootstrap:
+	$(MAKE) docker-build
+	docker compose up -d --remove-orphans
+	./docker/await_service_healthy.sh ssas
```
laurenkrugen-navapbc, line 63: is there a chance that we run into an issue where the container is ready but the api itself is not? worthwhile to also hit one of the `metadata` type endpoints to verify ssas is ready?

Makefile
michaeljvaldes: I'm glad you asked that! You're right -- our unit tests and integration tests are fundamentally different in that the unit tests operate independently in their own ecosystem, while the integration tests operate in the app network.
I split our integration tests into a separate compose file, which "includes" the main compose file. The integration tests now have the ssas service as a dependency, so simply running either of the tests will start up the required services automatically.

Makefile
michaeljvaldes: Actually that's basically what's happening!
The `await_service_healthy.sh` script waits for the ssas service's docker health check to declare that it is healthy. The health check itself pings ssas's `/_health` endpoint -- just like the load balancers in our deployed environments!

docker-compose.yml
michaeljvaldes: Ooh good point -- I reduced the health check interval to 40s.

Makefile
michaeljvaldes: It's very unlikely due to the short reset window and long health check interval, but I think you've pointed out an interesting issue with the order of the commands. In the `docker-bootstrap` recipe, I've moved `load-fixtures` to occur before starting ssas and waiting for it to be healthy. I think that addresses the root of the issue!

compose.integration-test.yml
```diff
@@ -0,0 +1,22 @@
+include:
```
carlpartridge, line 1: Do you think it would be reasonable to pare the name down of all the sub docker compose files to something like `dc.integration-tests.yml`?

docker-compose.migrate.yml
carlpartridge: Oops I did miss that...  However if it just starts up a single service (migrate) why cant we just use a docker image/docker run?

compose.integration-test.yml
michaeljvaldes: How about `compose.integration-test.yml` ?
I'd rename the others too (compose.yml, compose.migrate.yml, compose.test.yml)

ops/migrations_test.sh
```diff
@@ -18,8 +18,7 @@ usql $DB_HOST_URL -c 'create database ssas_test;'
 
 echo "Running SSAS migration tests"
 
-cd db/migrations
-DATABASE_URL=$TEST_DB_URL DEBUG=true go test -tags=migrations -v
+DATABASE_URL=$TEST_DB_URL go test -tags=migrations ./db/migrations
```
carlpartridge, line 21: Likely outside the scope of this ticket but do we think these migration tests are useful?  My loose understanding is that they are mostly designed to verify up and down for every migration.  Which seems a bit excessive?

compose.integration-test.yml
carlpartridge: Works for me!

ops/migrations_test.sh
michaeljvaldes: I agree it's outside the scope of this ticket, but I do not think they are useful.

docker-compose.migrate.yml
michaeljvaldes: Good point, I've removed it in favor of a docker run command.

Makefile
michaeljvaldes: Correction: the unit tests actually do depend on the `db` service in the main compose file, they just set up a separate database within that same container. I could go one of two ways here: continue to use that same container and "include" the main db service as a dependency of the unit tests, or give the unit tests their own db service. 
I opted for the latter so that the unit tests are fully independent, but I could be convinced in either direction.

carlpartridge: Agreed, latter sounds preferable.

michaeljvaldes: Fixed!

carlpartridge (dismissed):

juliareynolds-nava (dismissed):

bhagatparwinder (dismissed):

carlpartridge (approved):
Nice, this is great!  Looks like bake just builds services in parallel?

bhagatparwinder (approved):

PR #328, opened 2026-06-22
BCDA-10099: token rate limiting

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10099

## 🛠 Changes

- Config changes
  - Added `RateLimitThreshold` and `RateLimitDurationMinutes`
  - Updated env files
  - Updated compose file
- Added rate-limiting middleware
  - Added two caches `clientIDToACOIDCache` and `rateLimitCache`
  - Getting ACO ID from System
  - Add Retry-After header (dynamic)
  - Check for race conditions when multiple concurrent requests
  - Check for negative caching
- Added middleware to router
- Unit testing 

## ℹ️ Context

The v1 /token endpoint had no request throttling. A high-volume request attack/abuse could easily exhaust the database connection pool. We have implemented a configurable rate limiter for the /token endpoint to protect the authentication service and database from load spikes (reduce accidental abuse and denial of service).

## 🧪 Validation

Added unit/integration testing


ssas/cfg/configs/opensbx.env
michaeljvaldes, line 1: Nice catch!

ssas/service/public/middleware.go
```diff
@@ -181,3 +187,110 @@ func contains(list []string, target string) bool {
...
+	duration := time.Duration(cfg.RateLimitDurationMinutes) * time.Minute
+	// cache for rate limiting each aco based on aco ID
+	rateLimitCache = cache.New(duration, 1*time.Minute)
+}
+
+type groupXData struct {
+	CMSIDs []string `json:"cms_ids"`
+}
+
+func getACOIDFromSystem(ctx context.Context, system ssas.System, gr ssas.GroupRepository) (string, error) {
```
michaeljvaldes: This function seems really generally useful -- do you think it's worth moving this to a public function in, say, systems.go?

ssas/service/public/middleware.go
```diff
@@ -181,3 +187,110 @@ func contains(list []string, target string) bool {
...
+		if err != nil {
+			// Key not found in cache (first request in window). Initialize to 1.
+			rateLimitCache.Set(acoID, 1, duration)
+			newVal = 1
+		}
+
+		if newVal > cfg.RateLimitThreshold {
+			logger.Warnf("Rate limit exceeded for ACO %s (Client %s): %d requests in %v", acoID, clientID, newVal, duration)
+			w.Header().Set("Retry-After", "300") // 5 minutes retry recommendation
+			http.Error(w, "Too Many Requests", http.StatusTooManyRequests)
```
michaeljvaldes, line 267: I know there was some discussion around how much info to put in our response. 
@greg-pf is there anything specific we need to include in the message?

ssas/service/public/middleware.go
```diff
@@ -181,3 +187,110 @@ func contains(list []string, target string) bool {
 	}
 	return false
 }
+
+var (
```
michaeljvaldes: Does it make sense to make these caches fields of the PublicMiddlewareHandler instead of package-level vars? Or maybe move them to a different file/service and make that service a dependency of PublicMiddlewareHandler?

ssas/service/public/middleware.go
bhagatparwinder: I think if we want it within certain domain boundaries, it could go into `groups.go`. I know it takes System as an argument, but we query the Group's metadata. The helper also relies on `GroupRepository`. If I put it under `systems.go` we would introduce dependency on `groups` inside the system domain. I think we have two options:
1. Move to `groups.go` as a public helper function.
2. Keep it local as is (as currently it is only consumed here, so we keep the entity models and repos clean of domain-specific parsing).

Let me know your thoughts.

ssas/service/public/middleware.go
bhagatparwinder: I like the idea of moving caches to fields on the `publicMiddlewareHandler`. Cleanup test issues and state stays localized to the middleware instance.

ssas/service/public/middleware_test.go
```diff
@@ -279,3 +280,146 @@ func (s *PublicMiddlewareTestSuite) TestContains() {
...
+
+	creds2, err := s.h.sr.RegisterSystem(context.Background(), "Limit Test 2", groupID2, cfg.DefaultScope, "", []string{}, "tracking-limit-id-2")
+	require.Nil(s.T(), err)
+
+	reqDiff := httptest.NewRequest("POST", "/token", nil)
+	reqDiff.SetBasicAuth(creds2.ClientID, creds2.ClientSecret)
+	rrDiff := httptest.NewRecorder()
+
+	rateLimitedHandler.ServeHTTP(rrDiff, reqDiff)
+	assert.Equal(s.T(), http.StatusOK, rrDiff.Code, "Request for a different ACO should succeed")
```
carlpartridge, line 344: Im not totally understanding how the cache gets decremented after a certain amount of time.  As such I think its necessary to get a test that verifies that after the timeout duration has passed that another request succeeds again.  I think this could be done by converting the timeout config from minutes to seconds, then we can set it to 3s, then sleep for 3s at the end of this, then send a new request.

ssas/cfg/configs/dev.env
```diff
@@ -4,4 +4,6 @@ SSAS_DEFAULT_SYSTEM_SCOPE=bcda-api
 SSAS_IDLE_TIMEOUT=120
 SSAS_LOG=/var/log/api/ssas.log
 SSAS_READ_TIMEOUT=10
-SSAS_WRITE_TIMEOUT=20
\ No newline at end of file
+SSAS_WRITE_TIMEOUT=20
+SSAS_RATE_LIMIT_THRESHOLD=100
```
carlpartridge: 100 Seems like far too much as the default.  I believe the goal of this ticket is to just implement it but not have it be restrictive?  If so then this is probably good enough.  Assuming we have 4 instances running in prod this would be the equivalent of 400 requests every 5m, at least as far as I understand how a per instance cache would handle this.

ssas/service/public/middleware.go
carlpartridge: I think I would put it in groups.go as well.

ssas/service/public/middleware.go
carlpartridge: This would also be handled by BCDA API right?  Do we already watch for 429s in the API route function that calls this SSAS route?  If so do we need to distinguish between this issue and other 429s and return specific error messages for each?

ssas/service/public/middleware.go
michaeljvaldes: I like groups.go too!

ssas/service/public/middleware.go
michaeljvaldes: Ah that makes sense.

ssas/service/public/middleware.go
bhagatparwinder: Done

ssas/service/public/middleware.go
bhagatparwinder: Will take care of it in bcda-app

ssas/cfg/configs/dev.env
bhagatparwinder: Correct. I have it high enough that it doesn't throttle at the moment. It is 400/5 minutes or 4800 in an hour. So this currently allows everyone to go through until we communicate and are ready to implement this (lower this limit).

ssas/service/public/middleware.go
bhagatparwinder: Done!

ssas/service/public/middleware_test.go
bhagatparwinder: The cache doesn't have a decrement at all; it is a fixed-window cache (based on time). We are not limiting ACOs just by the number of requests they can make, but also the time during which that number of requests is allowed.

We create a counter with a set TTL (currently with 5 minutes). The counter starts from 1 and could go to 100 before 429 happens. If the counter expiration is passed, that counter itself is recreated with a value of 1 (ACO has a new lease on requests they can make).

If > 100, we handle sending 429. If past 5 minutes, `go-cache` considers the key expired, and the counter is gone.

If you check `h.rateLimitCache.Add(acoID, 1, duration)`. It creates the key for the client and sets the value of 1 and the expiration to 5 minutes.

bhagatparwinder: We use `go-cache` already in the same way for the token denylist.

```dl.c = cache.New(defaultCacheTimeout, cfg.cacheCleanupInterval)```

carlpartridge: I guess the issue is that the 5m window is constantly moving right?  Like we dont get all requests instantly, we get them over a period of time.  Eg (assuming 5 max before we return 429s in a 5m window) if we have 2 requests at minute 0, 2 requests at minute 2, and 2 requests at minute 4, then this sixth request at minute 4 I would expect to result in a 429 (2+2+2 is one over in a 5m window).  However at minute 5.1 we are down to just 4 requests in the past 5m so we can allow 1 more request through.

Or maybe this is just unnecessary complexity?  I guess its roughly equivalent to say at the very first request, start the cache up at 1, start the 5m timer, and within the next 5m you get x more total requests.  Total reset after 5m as judged by the very first request.  Ok I think Ive talked myself down 😅.

carlpartridge (dismissed):

ssas/service/public/middleware_test.go
michaeljvaldes: I agree, the implementation details aren't _exactly_ what I had imagined, but I think it's a clever use of the cache library features and it should be similarly effective at preventing too many token requests.

michaeljvaldes (dismissed):

ssas/service/public/middleware_test.go
bhagatparwinder: I have to clarify, the rate-limit window is **fixed** (anchored to the time of the first request). It is not rolling. So it is fairly simple.

You make a request 1 at 12 PM, we have a budget of 100 (or 400 requests based on 4 instances) for you till 12:05. This repeats again once the window expires or you have gone over the limit.

Rolling window would be a fair bit of complexity.

michaeljvaldes (approved):

carlpartridge (approved):

